Issue #86, opened 2014-01-07
Harmonize Date/Time format with Node.js in utimes/futimes


humphd:
http://stackoverflow.com/questions/20876407/what-format-is-used-for-nodejs-fs-utimes/20876502#20876502

Node.js allows for either Unix Timestamps or JavaScript Dates, and converts between the two.  We should do the same.  We currently assume only JavaScript Dates.

humphd:
See also https://nodejs.org/api/fs.html#fs_stat_time_values.

humphd:
https://github.com/streamich/memfs/blob/master/src/volume.ts#L477-L493 is a public domain implementation if someone wants to hack this into filer.

humphd:
Fixed in https://github.com/filerjs/filer/commit/443b58203858170fe50c38debd6343cb082686ec

PR #168, opened 2014-03-29
Issue #86: Added support for Unix Timestamps

kwkofler:


humphd:
Fixed this up so it could land on `master`, updated our `Stats` implementation to match node too, so we have both Date and timestamp versions of all times, fixed docs to match, fixed a ton of tests, added new tests.

https://github.com/filerjs/filer/commit/443b58203858170fe50c38debd6343cb082686ec.

Issue #122, opened 2014-02-20
Implement sh.mv()


humphd:
We need an implementation of the `mv` command on `FileSystemShell`, with tests and docs in the README. It should work in similar ways to the unix `mv` command, and do the right thing with files vs. dirs.

sedge:
@modeswitch The `mv` utility will automatically overwrite files during the move unless told not to. Do we want to keep this for our version?

modeswitch:
@sedge Yes, that sounds reasonable.

PR #170, opened 2014-03-30
Issue #122 - Added sh.mv with tests and documentation

kwkofler:
New request -_- Sorry about the last one.

### DIFF
--- a/src/shell/shell.js
+++ b/src/shell/shell.js
@@ -499,19 +499,32 @@ define(function(require) {
     callback = callback || function(){};
 
     if(!path) {
-      callback(new Errors.EINVAL('missing path argument'));
+      callback(new Errors.EINVAL("Missing path argument"));
       return;
     }
-    else if (path === '/'){
+    else if (path === '/') {
       callback();
       return;
     }
-    function _mkdirp(path, callback){
+    function _mkdirp(path, callback) {
       fs.stat(path, function (err, stat) {
-        //doesn't exist
-        if (err && err.code === 'ENOENT') {
+        if(stat) {
+          if(stat.isDirectory()) {
+            callback();
+            return;
+          }
+          else if (stat.isFile()) {
+            callback(new Errors.ENOTDIR());
+            return;
+          }
+        }
+        else if (err && err.code !== 'ENOENT') {
+          callback(err);
+          return;
+        }
+        else {
           var parent = Path.dirname(path);
-          if(parent === '/'){ //parent is root
+          if(parent === '/') {
             fs.mkdir(path, function (err) {
               if (err && err.code != 'EEXIST') {
                 callback(err);
@@ -521,10 +534,10 @@ define(function(require) {
               return;
             });
           }
-          else { //parent is not root, make parent first
+          else {
             _mkdirp(parent, function (err) {
               if (err) return callback(err);
-              fs.mkdir(path, function (err) { //then make dir
+              fs.mkdir(path, function (err) {
                 if (err && err.code != 'EEXIST') {
                   callback(err);
                   return;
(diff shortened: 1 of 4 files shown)


sedge:
This is a pretty good patch!

@modeswitch Is this approach to `mv` what we want? I'm fairly sure `fs.rename` can do the majority of the work for us though I'm still investigating. I need this for #273 , so if `rename` isn't enough I'll probably update this patch instead of the one I'm developing.

modeswitch:
I had a quick look at it seems reasonable. The code probably needs to be updated to bring it in line with recent refactoring.

Can you give it a review (adding comments to the patch) and we can discuss any changes you feel are required. If most of the work is done already, let's use this.

src/shell/shell.js
```diff
@@ -337,6 +338,139 @@ define(function(require) {
...
+      return;
+    }
+
+    function move(sourcepath, destpath, callback) {
+      sourcepath = Path.resolve(this.cwd, sourcepath);
+      destpath = Path.resolve(this.cwd, destpath);
+      var destdir = Path.dirname(destpath);
+
+      // Recursively create any directories on the destination path which do not exist
+      shell.mkdirp(destdir, function(error) {
```
sedge: The `mv` utility won't ensure the path is correct by creating folders. Instead, it errors if the destination directory doesn't exist.

src/shell/shell.js
```diff
@@ -337,6 +338,139 @@ define(function(require) {
...
+        if(error) {
+          callback(error);
+          return;
+        }
+
+        // If there is no node at the source path, error and quit
+        fs.lstat(sourcepath, function(error, sourcestats) {
```
sedge: This is identical to how `fs.rename` behaves.

src/shell/shell.js
```diff
@@ -337,6 +338,139 @@ define(function(require) {
...
+          if(error) {
+            callback(error);
+            return;
+          }
+
+          fs.lstat(destpath, function(error, deststats) {
```
sedge: The more important check would be, *does the parent directory exist*. This is how `fs.rename` behaves.

sedge: In addition, there's a difference between a destination of `/folder` and `/folder/`. In the first case, `source` must also be a folder and will completely blow away `/folder` and it's contents. This line assumes the second case.

tests/spec/shell/mv.spec.js
```diff
@@ -0,0 +1,488 @@
+define(["Filer", "util"], function(Filer, util) {
```
sedge: This is missing a test for failure on a *source/folder* to *destination/file* move.

src/shell/shell.js
```diff
@@ -337,6 +338,139 @@ define(function(require) {
...
+              callback(error);
+              return;
+            }
+            
+            if(deststats) {
+              // If the destination is a directory, new destination is destpath/source.basename
+              if(deststats.isDirectory()) {
+                destpath = Path.join(destpath, Path.basename(sourcepath));
+              }
+            }
```
sedge: This needs some extra logic to prevent errors in the case of an attempted `folder` to `file` move.

src/shell/shell.js
```diff
@@ -337,6 +338,139 @@ define(function(require) {
...
+                return;
+              }
+              // If the source is a file, link it to destination and remove the source, then done
+              if(sourcestats.isFile() || sourcestats.isSymbolicLink()) {
+                fs.link(sourcepath, destpath, function(error) {
+                  if (error) {
+                    callback(error);
+                    return;
+                  }
+                  shell.rm(sourcepath, {recursive:true}, function(error) {
```
sedge: @modeswitch Should this be an `fs.unlink()`?

sedge:
@modeswitch The `mv` utility can accept multiple source files to be moved to a directory. Do we want to support this?

sedge:
The new version of this PR is #346

sedge:
The [POSIX standard](http://pubs.opengroup.org/onlinepubs/009604599/utilities/mv.html) says that the `mv` utility relies on `rename()`, except for cross-filesystem moves. At the moment, this patch is doing the `rename()` work manually, but it's working. Should I switch it to use `fs.rename()`?

modeswitch:
This can't be automerged. Can you investigate?

sedge:
@modeswitch Here's the updated PR: https://github.com/filerjs/filer/pull/346

It's missing a test case, which shouldn't be too difficult to add. Other than that I believe it's ready for review.

PR #172, opened 2014-04-17
Issue #158 - Added test to ensure that watchers monitoring hardlinks are...

kwkofler:
... notified when the original file is updated

I'm not quite sure what to do with this, so I decided to open a pull request anyway. I've created the test, and it works, but the behavior that it's testing does not. Hardlink watchers will time out when waiting for alerts to changes in the original file. Should I file this as a new bug?


humphd:
Finishing this in https://github.com/filerjs/filer/pull/595.

PR #176, opened 2014-04-18
Issue 87 - Added fs.fsync() as a no-op, and added tests. Modified docume...

kwkofler:
...ntation.


humphd:
Taking this over the line in https://github.com/filerjs/filer/pull/593.  Sorry these got stalled @kwkofler, thanks for doing them.

Issue #42, opened 2013-11-27
Support legacy arg layout for fs.read() in node.js


humphd:
The `read()` method on node's `fs` can be called one of two ways, and we don't currently support the old way (from https://github.com/joyent/node/blob/master/lib/fs.js#L413-L438):

``` javascript
fs.read = function(fd, buffer, offset, length, position, callback) {
  if (!util.isBuffer(buffer)) {
    // legacy string interface (fd, length, position, encoding, callback)
    var cb = arguments[4],
        encoding = arguments[3];

    assertEncoding(encoding);

    position = arguments[2];
    length = arguments[1];
    buffer = new Buffer(length);
    offset = 0;

    callback = function(err, bytesRead) {
      if (!cb) return;

      var str = (bytesRead > 0) ? buffer.toString(encoding, 0, bytesRead) : '';

      (cb)(err, str, bytesRead);
    };
  }

  function wrapper(err, bytesRead) {
    // Retain a reference to buffer so that it can't be GC'ed too soon.
    callback && callback(err, bytesRead || 0, buffer);
  }

  binding.read(fd, buffer, offset, length, position, wrapper);
};
```

PR #179, opened 2014-04-18
Issue #42 - Added support for node.js' legacy arg layout for fs.read()

kwkofler:


humphd:
Looks like node has dropped this, closing.

PR #180, opened 2014-04-18
Fixes issue 75

pbouianov:
As the array flags were already reflected in the README documentation
and example, this only required a test


humphd:
Rebased and merged in https://github.com/filerjs/filer/commit/1775c24d37cdc36b634d8dbc68f466d8426687fd.  Sorry it took so long @pbouianov and thanks for this work!

PR #336, opened 2014-12-08
Fix #273 - Configure default builtins and expose FileSystem API on shell instances

### LINKED ISSUE #273: Configure default builtins and expose FileSystem API on shell instances
Package them as optional programs that can run from the fs using `exec`.

sedge:
This patch has removed a couple of default builtins and created a place to store optional ones. This optional builtin system keeps the builtins in one place, and makes it easy to create new shell repos including any combination of them.

All that's left is implementing and testing the `mv` and `cp` shell builtins.


sedge:
@modeswitch I need to add the `mv` and `cp` commands to the shell, plus associated tests.

sedge:
@modeswitch I'm going to split the `mv` and `cp` stuff into separate issues. This is ready for review.

tests/spec/shell/cat.spec.js
```diff
@@ -1,20 +1,25 @@
 var Filer = require('../../..');
 var util = require('../../lib/test-utils.js');
 var expect = require('chai').expect;
+var optionals = require('../../../src/shell/optional');
 
-describe('FileSystemShell.cat', function() {
+describe('(optional) FileSystemShell.cat', function() {
```
modeswitch: Why mark these as optional? I'm don't see how that improves the test reporting, and it may be confusing.

src/shell/optional.js
```diff
@@ -0,0 +1,116 @@
...
+optional.touch = function(sh, path, options, callback) {
+  touch.call(sh, path, options, callback);
+};
+
+optional.cat = function(sh, files, callback) {
+  cat.call(sh, files, callback);
+};
+
+/**
+ * Access to optional methods for shell
```
modeswitch: Can you explain this in more detail?

modeswitch:
@sedge I've added a couple of comments, but I don't have time to do a thorough review today. I'll update it again this weekend.

tests/spec/shell/cat.spec.js
sedge: I'll pull them. It's more a sign of my confusion about where the optional stuff will live when we split the shells off.

src/shell/optional.js
sedge: We should probably chat when you have some time to focus my approach to the problem of how to provide access to optional shell methods. This was intended to allow something like:

```javascript
var filer = require('filer');
var optionals = filer.optionals;
var sh = filer.Shell;

optionals.bindTouch(sh);

module.exports = sh;
```

This example is at best pseudocoded, and my approach hasn't really solved this problem yet. I thought it might be a good starting point for discussion though!

modeswitch: Yep, let's set up a time to chat.

humphd:
We'll do this, but in another way.  Going to close for now.  Thanks @sedge.

PR #345, opened 2015-01-14
[WIP] Fix 122 - Implement the mv command

sedge:
This is an alternate approach the PR #122 and is still being developed. Namely, the tests are failing (predictably) because the method is still being implemented.


sedge:
#122 is a better approach.

PR #346, opened 2015-01-15
Fix #122: Add mv shell utility + tests + docs

### LINKED ISSUE #122: Implement sh.mv()

sedge:
This is an adaptation of #170.

### DIFF
--- a/tests/spec/shell/mv.spec.js
+++ b/tests/spec/shell/mv.spec.js
@@ -1,0 +1,488 @@
+define(["Filer", "util"], function(Filer, util) {
+  
+  describe('FileSystemShell.mv', function() {
+    beforeEach(util.setup);
+    afterEach(util.cleanup);
+
+    it('should be a function', function() {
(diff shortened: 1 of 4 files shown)


tests/spec/shell/mv.spec.js
```diff
@@ -0,0 +1,342 @@
+var Filer = require('../../..');
+var util = require('../../lib/test-utils.js');
+var expect = require('chai').expect;
+
+describe('FileSystemShell.mv', function() {
```
sedge: I forgot, but this also needs a test for failure on any of `destination`'s parent directories not existing.

sedge:
To reiterate, this implementation of `mv` doesn't allow the movement of multiple files in one command. If we want to include that it'll need one more revision.

Issue #105, opened 2014-02-06
Deal with trailing slashes on path names like node.js


humphd:
Via Simon Wex, Node.js seems to know how to deal with a trailing slash on a directory path, where we want to include it in the actual name:

``` javascript
> require('fs').readdir('src', function(err, r){ console.log(err, r); });
> null [ 'adapters',
  'constants.js',
  'error.js',
  'fs.js',
  'index.js',
  'path.js',
  'providers',
  'shared.js' ]
```

vs

``` javascript
> require('fs').readdir('src/', function(err, r){ console.log(err, r) });
> null [ 'adapters',
  'constants.js',
  'error.js',
  'fs.js',
  'index.js',
  'path.js',
  'providers',
  'shared.js' ]
```

PR #351, opened 2015-05-12
fs.rename() should normalize paths before using, dir vs. dir/

humphd:
We hit a bug in Brackets where renames are using dir paths like this: `fs.rename("/old-path/", "/new-path/", ...)` and Filer is using the trailing slash as part of the path.  We fixed something like this in issue #105 but seem to have missed the rename case.

### DIFF
--- a/src/filesystem/implementation.js
+++ b/src/filesystem/implementation.js
@@ -1976,6 +1976,9 @@ function rename(fs, context, oldpath, newpath, callback) {
   if(!pathCheck(oldpath, callback)) return;
   if(!pathCheck(newpath, callback)) return;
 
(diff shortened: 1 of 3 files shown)


gideonthomas:
r+ looks great

humphd:
@modeswitch I need a `grunt publish` to get this change on npm.  Can you do that please?

modeswitch:
@humphd `grunt publish` appears to be broken in Node v0.12. I'll need to investigate when I have a sec, unless @sedge has time.

humphd:
Lovely :(  fwiw, I'm not blocked on this, since we're using it via submodule.  But it would be nice if it was uptodate on npm too.

PR #355, opened 2015-06-01
Fixed #354 - Add standalone Path and filer-buffer modules

### LINKED ISSUE #354: Automate creation of dist/filer-buffer.js and dist/path.js
In Bramble I need the `Filer.Buffer` code and `Filer.Path`, but I don't want the rest of Filer (i.e., it will live in a remote window that I talk to over `MessageChannel`).  For download size, I really want to be able to do the following in Brackets:

```js
var Path = require("thirdparty/filer/dist/path");
var Buffer = require("thirdparty/filer/dist/filer-buffer");
```
Then I don't have to do:
```js
var Filer = require("thirdparty/filer/dist/filer.min");
var Path = Filer.Path;
var Buffer = Filer.Buffer;
``` 
This will be a huge download saving.

sedge:


humphd:
r+ with travis.  @gideonthomas can you r?

gideonthomas:
nit: I don't think we need to prefix `buffer` as `filer-buffer` if we're not doing it for `path`. Either way, r+

PR #356, opened 2015-06-01
Fixed #352 - Removed `process` from browserified version of filer

### LINKED ISSUE #352: Remove Browserify's `process` from built filer, stop process-tick events
Filer floods `postMessage` with `process-tick` events, simulating Node's `nextTick` stuff.  We dont' need it, and it's just there because Browserify is adding it:

```js
},{"FWaASH":10}],10:[function(_dereq_,module,exports){
// shim for using process in browser

var process = module.exports = {};

process.nextTick = (function () {
    var canSetImmediate = typeof window !== 'undefined'
    && window.setImmediate;
    var canPost = typeof window !== 'undefined'
    && window.postMessage && window.addEventListener
    ;

    if (canSetImmediate) {
        return function (f) { return window.setImmediate(f) };
    }

    if (canPost) {
        var queue = [];
        window.addEventListener('message', function (ev) {
            var source = ev.source;
            if ((source === window || source === null) && ev.data === 'process-tick') {
                ev.stopPropagation();
                if (queue.length > 0) {
                    var fn = queue.shift();
                    fn();
                }
            }
        }, true);

        return function nextTick(fn) {
            queue.push(fn);
            window.postMessage('process-tick', '*');
        };
    }

    return function nextTick(fn) {
        setTimeout(fn, 0);
    };
})();

process.title = 'browser';
process.browser = true;
process.env = {};
process.argv = [];

function noop() {}

process.on = noop;
process.addListener = noop;
process.once = noop;
process.off = noop;
process.removeListener = noop;
process.removeAllListeners = noop;
process.emit = noop;

process.binding = function (name) {
    throw new Error('process.binding is not supported');
}

// TODO(shtylman)
process.cwd = function () { return '/' };
process.chdir = function (dir) {
    throw new Error('process.chdir is not supported');
};
```
We need it ripped out of our build step so it doesn't get added to the built filer.

sedge:
This patch also restricted the included builtins to what filerjs needs to pass its tests. I can pull those changes out if a separate patch would be better.


humphd:
Can you please run the tests in node, FF, and Chrome, and r+ if they all pass.  Nice work.  Please also update `dist/` before you land, and then update Bramble.

PR #359, opened 2015-06-29
Fix #357 - Path.resolve() should not crash with missing Path.relative()

### LINKED ISSUE #357: Path.resolve does not work
Try:
```js
Path.relative("/mydir", "/mydir/file");
```

Crashes with:
`TypeError: Object #<Object> has no method 'resolve'`

humphd:


gideonthomas:
tested this and it works! nice!
r+

humphd:
I've tested this in node, FF, Chrome, and Safari, and all are working now.  I had to update a few things to get our build to work again, but it's fine with these changes.

PR #370, opened 2017-05-21
Use transaction per operation in indexeddb.js, fix broken async tests in fs.stats.spec.js

humphd:
Well, after a long absence, in which I thought this code was perfect, I'm back with a fix :)  Our IndexedDB provider code has been broken forever, and I'm impressed it's worked at all.  We're getting reports in Thimble of users having corrupt databases.  I think one factor that might be adding to it is that we are currently holding a lifetime reference to a single `transaction` and `objectStore` instead of creating one per operation.

I might add a new provider, or switch out this one, to use one of the many well-maintained IndexedDB wrapper libs, which have better error handling and cross-browser support than we do here.  In the meantime, this fixes things.

Making this switch exposed some broken tests, where we were doing async operations, but ending our tests immediately as though they were sync.  Doing so caused the transaction to abort right away, which is good.  I've fixed those tests to make them properly async in our test suite.

I ran the tests in browser for all of:

* Chrome on macOS
* Firefox on macOS
* Safari on macOS
* Opera on macOS
* MS Edge on Windows 10
* IE11 on Windows 10
* Chrome on ChromeOS

Everything works well, with the odd instance of a timeout in `sh.ls and deep directory trees`, which is known to fail sometimes with the current test timeout period, so it's ignorable.  I'm going to try plugging this into Thimble and test that, too.


humphd:
Tested in Thimble (e.g., updated submodule in Bramble) and this seems to work well.

modeswitch:
> I think one factor that might be adding to it is that we are currently holding a lifetime reference to a single transaction and objectStore instead of creating one per operation.

This shouldn't be the case, and if so is definitely a bug. One thing to keep in mind is that IDB doesn't have explicit transaction commits. Each time we create a `context` from the `provider`, that should be creating a new transaction that is implicitly committed when no more IDB operations are added from inside a transaction callback.

Relevant code is here:
https://github.com/filerjs/filer/blob/6d3cec89ee5074f8393177bc10ae4d1c2440f12b/src/providers/indexeddb.js#L138 and https://github.com/filerjs/filer/blob/6d3cec89ee5074f8393177bc10ae4d1c2440f12b/src/filesystem/interface.js#L223

That said, I wouldn't be surprised if there was a bug :/

src/providers/indexeddb.js
```diff
@@ -10,67 +10,90 @@ var indexedDB = global.indexedDB       ||
...
-  } catch(e) {
-    callback(e);
+  var builder = new TransactionBuilder(this.db, this.mode);
+  builder.build();
+  if(builder.error) {
+    return callback(builder.error);
   }
+
+  var transaction = builder.transaction;
+  var objectStore = transaction.objectStore(FILE_STORE_NAME);
```
humphd: Indeed we should, confirmed here https://twitter.com/asutherland/status/866498053519192064.  I spent a long time reading other IDB wrapper libraries, and they are all over the map in terms of what they do with `InvalidStateError` handling--most do nothing.  I'll fix.

src/providers/indexeddb.js
```diff
@@ -10,67 +10,90 @@ var indexedDB = global.indexedDB       ||
...
+
+  var request = objectStore.clear();
+  request.onsuccess = function(event) {
+    callback();
+  };
+  request.onerror = callback;
```
humphd: Good catch, I'll fix.

src/providers/indexeddb.js
```diff
@@ -10,67 +10,90 @@ var indexedDB = global.indexedDB       ||
...
-      var result = event.target.result;
-      callback(null, result);
-    };
-    request.onerror = function onerror(error) {
-      callback(error);
-    };
-  } catch(e) {
-    callback(e);
+IndexedDBContext.prototype._put = function(key, value, callback) {
+  var builder = new TransactionBuilder(this.db, this.mode);
```
humphd: Yeah, I'l refactor.

humphd:
@modeswitch the bug is mine (i.e., in my idb implementation), and our code abstractions do the right thing.  I just need to do this fix to make it consistent.  Because all `fs` operations are assumed to be independent (e.g., we don't batch them like you might if you were reading inserting a ton of rows in a DB with pure SQL), we need to do transaction-per-operation.  It's amazing that it's worked this long without it!

humphd:
I did some more work on this.  First, a clarification: I was wrong in my initial assessment of this bug, and we are *not* holding a single transaction for the entire program.  Instead, we currently create a transaction per filesystem operation (e.g., `writeFile` or `stat`).  With this transaction, we then do any number of `get` or `put` operations on the DB, in some test cases as many as 208 (most operations do 5-20).  A disadvantage of this approach is that we are never able to use `READ_ONLY` transactions, since we almost always need to write metadata to a node (e.g., access time updates).

In this PR I've changed things so that we do a transaction per `put`, `get`, `delete`, or `clear` operation vs. per context.  This means that we can now use `READ_ONLY` along with `READ_WRITE` transactions, giving the underlying database a chance to parallelize some operations (only a single `READ_WRITE` transaction can happen at once).  This is great for cases like `fs.readdir`, where you could have a ton of parallel reads to do.  I've also dealt with situations where `indexeddb` operations can throw, which I [blogged about here](http://blog.humphd.org/when-in-doubt-webidl/).

Out of interest, I re-ran the test suite against as many browsers as I could (most are on my iMac with macOS, and MS browsers on Windows 10): 

| Browser | Passes Tests | Time (ms) |
|---------|---------------|-----------|
|Chrome  | ✅                     | 48.68 |
| Chrome Incognito | ✅  | 15.16 |
| Chrome (ChromeOS) | ✅  | 18.30 |
|Firefox | ✅  | 47.91 |
| Firefox Private Window | ❌  | N/A |
| Safari | ✅  | 92.41 |
| Safari Private Window | ❌  | N/A |
| Safari Beta Preview | ✅  | 87.13 |
| Opera | ✅  | 20.49 |
| Opera Private Window | ✅  | 11.02 |
| MS Edge | ✅  | 30.37 |
| MS Edge Private Window | ✅  | 4.63 |
| MS IE11 | ✅  | 30.21 |
| MS IE11 Private Window | ✅  | 4.57 |

Our test suite isn't a perfect test of performance, since we create/delete databases all the time vs. using just one; however, it's fascinating to see the different performance characteristics of these implementations.  The really fast ones must not be doing an `fsync` to write the data to disk.  If your browser crashed, I bet you'd lose that data.  Meanwhile, Safari is super slow, but I can hear it writing to disk for every operation.  Apple *really* doesn't want you to lose your data!

I think this change is a good one, but it's not as critical as I first feared.  @gideonthomas, can you take another look at this?  I could refactor this code more to eliminate code duplication, but I didn't because it will just make it harder to understand, without saving us that much space.

src/providers/indexeddb.js
```diff
@@ -11,66 +11,77 @@ var indexedDB = global.indexedDB       ||
                 global.msIndexedDB;
 
 function IndexedDBContext(db, mode) {
-  var transaction = db.transaction(FILE_STORE_NAME, mode);
-  this.objectStore = transaction.objectStore(FILE_STORE_NAME);
+  this.db = db;
 }
 
+IndexedDBContext.prototype._getObjectStore = function(mode) {
+  var transaction = this.db.transaction(FILE_STORE_NAME, mode);
```
Pomax: should `mode` have a fallback value?

tests/bugs/ls-depth-bug.js
```diff
@@ -25,7 +25,7 @@ describe('sh.ls and deep directory trees', function() {
         done();
       });
     });
-  });
+  }).timeout(15000);
```
Pomax: silly nit, but this 15 seconds is probably better maintained as a constant?

modeswitch:
I see a few problems with this approach:

1) Some operations must be executed atomically to avoid interleaving conflicting changes and leaving the file system in an inconsistent state (writing to a file, for example). By changing the context object to be non-transactional, it will introduce race conditions that I'm pretty sure we aren't testing for (and are, in general, difficult to test).

2) Some operations that are atomic could be made non-atomic (`readdir`, for example) in order to improve performance at the expense of consistency (you might see a resource in an inconsistent state, but the file system representation will be correct). This can be done using multiple contexts, where appropriate, instead of changing the underlying behavior of the context.

3) There is a bug (noted [here](https://github.com/filerjs/filer/blob/6d3cec89ee5074f8393177bc10ae4d1c2440f12b/src/providers/indexeddb.js#L139)) that causes read-only transactions to be executed differently (incorrectly?) on Chrome. I had a quick look and this may be fixed in Chrome now, but we should test it again to be sure.

Presently, the interface code is responsible for obtaining a context and passing it into the implementation (which separates the implementation from the provider code). Since the implementation code now needs to make decisions about how to split up work across transactions (contexts), the provider should be passed instead.

I hope the above makes sense (and is correct!), and I'm happy to jump on a call to discuss.

humphd:
@modeswitch I agree with you about the idea of leaving all operations for a context in a single transaction.  I debated whether to give up on what I've done here, and just leave it how we had it, but with some enhanced error handling.  I'm not super eager to refactor or rewrite the code in Filer, since it's been working wonderfully for years, and none of us has time to do major overhauls.  There's also not a ton of interest in other people, besides us, using it, so I don't think it's worth trying to do major rework.

I'll do another commit tomorrow that undoes the transaction-per-get/put stuff, and leave it at the level of the context.

Regarding the bug in Chrome about read-only transactions...I don't think it exists anymore.

src/providers/indexeddb.js
```diff
@@ -11,66 +11,80 @@ var indexedDB = global.indexedDB       ||
                 global.msIndexedDB;
 
 function IndexedDBContext(db, mode) {
-  var transaction = db.transaction(FILE_STORE_NAME, mode);
-  this.objectStore = transaction.objectStore(FILE_STORE_NAME);
+  var objectStore = null;
+  this._getObjectStore = function() {
```
gideonthomas: any reason you do this vs. attaching it to the prototype?

src/providers/indexeddb.js
```diff
@@ -11,18 +11,20 @@ var indexedDB = global.indexedDB       ||
...
-    var transaction = db.transaction(FILE_STORE_NAME, mode);
-    objectStore = transaction.objectStore(FILE_STORE_NAME);
-    return objectStore;
-  };
+  this.db = db;
+  this.mode = mode;
 }
 
+IndexedDBContext.prototype._getObjectStore = function() {
+  if(this.objectStore) {
```
Pomax: Would this make more sense as:
```
if (!this.objectstore) {
  // create object store
}
return this.objectstore;
```

PR #373, opened 2018-06-25
Update Filer to support Plan 9 virtio filesystem API in v86

humphd:
In order to get Filer working with v86 and allow a Linux VM to mount it via `virtio` Plan 9 Resource Sharing, I had to fix and extend it.  This PR upstreams those changes, which include:

* Add `chmod`, `fchmod`, `chown`, `fchown` along with tests.  There are more tests we could write later, but I've made sure to test what I used.  I've borrowed (and clearly marked) code related to this from node.js
* Added node's `fs.fsConstants` so people can get access to various mode bits
* Fixed a bug in `link_node` related to node info not being written
* Altered `symlink` to allow relative paths to match Unix behaviour (we also store an abspath so Filer still works as normal, but now we report the relpath)
* Removed an unnecessary FirefoxOS bug fix that was breaking WebTorront buffers with Filer
* Switched a bunch of `!=` and `==` to `!==` and `===` respectively
* Updated `README.md`
* Altered the node layout to include things needed by the above new fs methods.  I've also disambiguated `mode` vs. `type` in our implementation.
* add `package-lock.json`
* update all built files in `dist/` to reflect new code

I have not updated any deps, introduced new JS syntax, or tooling.  I'll do that in later PRs.

NOTE: this will be a breaking change (new node layout, new API additions), and we should reflect that in our npm version somehow.


tests/spec/fs.chmod.spec.js
```diff
@@ -0,0 +1,74 @@
...
+
+  it('should automatically set mode=755 for a directory', function(done) {
+    var fs = util.fs();
+
+    fs.mkdir('/dir', function(err) {
+      if(err) throw err;
+
+      fs.stat('/dir', function(err, stats) {
+        if(err) throw err;
+        expect(stats.mode & 0o755).to.equal(0o755);
```
modeswitch: Typo here? `0o755` -> `0x755`?

modeswitch: NM, it's supposed to be octal.

modeswitch: Not sure why the automated tests are failing. Assuming these ran correctly on your machine.

modeswitch: Oh, I see. Travis is using Node `0.10` to run our tests, and it doesn't understand octal numbers. We should switch to a newer version of Node. I'd suggest `8.11.3`, which is the current LTS build.

modeswitch: I've updated the Travis config on `develop`. If you do a rebase on your PR it should build and test cleanly now.

humphd: Good catch!  I've rebased to pick up the version change on Travis, and it's passing now.

PR #374, opened 2018-06-28
Switch to parceljs for bundling, eslint for linting, drop grunt

humphd:
This is a first pass at modernizing our bundling and build in order to support refactoring of the code to `es6`, etc.  It does the following:

* drops `grunt` in favour of a few `package.json` scripts.  NOTE: we need to replace the auto-deploy stuff on `npm`, but the right way to do that now is via [Travis npm deploys](https://docs.travis-ci.com/user/deployment/npm/).  We can do that in a follow-up PR.
* uses [Parcel.js](https://parceljs.org/getting_started.html) for bundling.  I also tried with rollup.js, and it gets complicated.  Parcel literally just works out of the box, and adds Babel support, uglify, etc.
* switched to `eslint` from `jshint`.  Have a look at the `.eslintrc.json` and see if you have changes you want.  NOTE: I have *not* run `--fix` to clean up all the errors it finds (380, mostly small things).  Once we're happy with the config, that can be done in a follow-up PR so this PR doesn't get huge.
* I've updated our dependencies and/or removed everything we don't need.
* I fixed a few other small things (typo in a test that was caught by updating `chai`), fixed the License SPDX format so `npm` won't complain.

In a follow-up PR, we can also add automatic Travis-based browser testing via Karma or Puppeteer or some other headless browser API.  I didn't want this to become a massive PR.  We can build on this in subsequent work.


humphd:
I need to rebase, my laptop was way out of date.  I'll push an updated commit soon...

humphd:
OK, rebased and fixed a few things:

* Removed `Babel` transpiling for now, since we don't use `import` in our tests.  When we switch everything to `es6` we can add it then and Parcel will just do it.
* Updated `minimatch` dep to latest
* Fixed a few code issues that caused tests to break, specifically, removing an old hack for Firefox OS Typed Arrays that isn't needed anymore (RIP FFxOS)

I'll need to rebase #373 on this before we land that too.

humphd:
I've added support for the `"browser"` field in `package.json`, which also involved updating our tests to `require()` the right thing in our tests.  Most of the changes in this commit are related to having Parcel ignore the `"browser"` entry in `package.json` and use` ./src/index.js` instead.  I've also removed unnecessary `require()`s for `Filer` where it wasn't being used.  Finally, I've dumped a few tests that really aren't needed anymore in 2018.

modeswitch:
Could you do a quick rebase on this to make sure everything is passing?

humphd:
OK, I rebased to pick up the new things you just landed.  I also updated Parcel (it was released twice again since I did my initial PR!).  I found a few small bugs that I've fixed now.

NOTE: the automated tests are not yet going to report properly in Travis, because it requires a browser.  For now, pull my branch and do `npm test` and then open your browser to the URL it gives.  After we land this, I'll do some more work to get Travis to run a browser headless.  This way we'll get proper IndexedDB results on Travis vs. our fake node memory provider we use now.  I also need to deal with all the eslint issues in follow ups.

PR #380, opened 2018-08-22
Adding support for promises. Closes #379, #382

### LINKED ISSUE #379: Support for the Promise API.
There is a new [promise](https://nodejs.org/api/fs.html#fs_fs_promises_api) api that converts every callback method into promise-style api. Once converted, filer will be able to support async/await as well.

dsych:
1. Attaching `promises` object on FileSystem similar to how node [implements](https://nodejs.org/api/fs.html#fs_fs_promises_api) it.
    * Adding deprecation warning to `exists` method, since it is obsolete in node, [see](https://nodejs.org/api/fs.html#fs_fs_exists_path_callback).
    * Promise is always rejected, when promise-based `exists` is called.
2. Attaching `promises` object on Shell that returns promise-based version of all method accept for `pwd`.

Simple test code to see how it works. 
> **Note:** I am using provider for [node](https://github.com/filerjs/filer-fs).
```js
const fs = new FileSystem({provider: new nodejs({name: guid(), keyPrefix: guid()})}, async (err, fs) => {
  const Shell = fs.Shell;
  const sh = new Shell().promises;
  fs = fs.promises;
  const fd = await fs.open('/myfile', 'w+');
  await fs.close(fd);
  const stats = await fs.stat('/myfile');
  console.log(stats);

  await sh.touch('/file1');
  console.log(await sh.ls('/'));
});
```

#379  #382

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,8 +110,8 @@ Like node.js, callbacks for methods that accept them are optional but suggested 
 you omit the callback, errors will be thrown as exceptions). The first callback parameter is
 reserved for passing errors. It will be `null` if no errors occurred and should always be checked.
 
-#### Support for Promises is here!
-Promise based API mimics the way Node [implements](https://nodejs.org/api/fs.html#fs_fs_promises_api) them. Both `Shell` and `FileSystem` now have `promises` object attached alongside regular callback style methods. Method names are identical to their callback counterparts with the difference that instead of receiving third argument as a callback, they return a Promise that is resolved or rejected based on the success of method execution.
+#### Support for Promises!
+The Promise based API mimics the way Node [implements](https://nodejs.org/api/fs.html#fs_fs_promises_api) them. Both `Shell` and `FileSystem` now have a `promises` object attached alongside the regular callback style methods. Method names are identical to their callback counterparts with the difference that instead of receiving a final argument as a callback, they return a Promise that is resolved or rejected based on the success of method execution.
 > Please note that `exists` method will always throw, since it was [deprecated](https://nodejs.org/api/fs.html#fs_fs_exists_path_callback) by Node.
 
 See example below:
--- a/README.md
+++ b/README.md
@@ -110,6 +110,21 @@ Like node.js, callbacks for methods that accept them are optional but suggested 
 you omit the callback, errors will be thrown as exceptions). The first callback parameter is
 reserved for passing errors. It will be `null` if no errors occurred and should always be checked.
 
+#### Support for Promises is here!
+Promise based API mimics the way Node [implements](https://nodejs.org/api/fs.html#fs_fs_promises_api) them. Both `Shell` and `FileSystem` now have `promises` object attached alongside regular callback style methods. Method names are identical to their callback counterparts with the difference that instead of receiving third argument as a callback, they return a Promise that is resolved or rejected based on the success of method execution.
+> Please note that `exists` method will always throw, since it was [deprecated](https://nodejs.org/api/fs.html#fs_fs_exists_path_callback) by Node.
+
+See example below:
+
+```javascript
+const fs = new Filer.FileSystem().promises;
+fs.open('/myfile', 'w+')
+  .then(fd => fs.close(fd))
+  .then(() => fs.stat('/myfile'))
+  .then(stats => { console.log(`stats: ${JSON.stringify(stats)}`); })
+  .catch(err => { console.error(err); })
+```
+
 #### Filer.FileSystem(options, callback) constructor
 
 File system constructor, invoked to open an existing file system or create a new one.
--- a/README.md
+++ b/README.md
@@ -80,6 +80,7 @@ instance.  By default, a new [IndexedDB](https://developer.mozilla.org/en/docs/I
 database is created for each file system. The file system can also use other
 backend storage providers, for example [WebSQL](http://en.wikipedia.org/wiki/Web_SQL_Database)
 or even RAM (i.e., for temporary storage). See the section on [Storage Providers](#providers).
+<a name="overviewExample"></a>
 
 ```javascript
 var fs = new Filer.FileSystem();
@@ -110,7 +111,7 @@ Like node.js, callbacks for methods that accept them are optional but suggested 
 you omit the callback, errors will be thrown as exceptions). The first callback parameter is
 reserved for passing errors. It will be `null` if no errors occurred and should always be checked.
 
-#### Support for Promises!
+#### Support for Promises
 The Promise based API mimics the way Node [implements](https://nodejs.org/api/fs.html#fs_fs_promises_api) them. Both `Shell` and `FileSystem` now have a `promises` object attached alongside the regular callback style methods. Method names are identical to their callback counterparts with the difference that instead of receiving a final argument as a callback, they return a Promise that is resolved or rejected based on the success of method execution.
 > Please note that `exists` method will always throw, since it was [deprecated](https://nodejs.org/api/fs.html#fs_fs_exists_path_callback) by Node.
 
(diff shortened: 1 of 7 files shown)


humphd:
We just landed my eslint fixes on `develop`, can you (or did you?) rebase?  Also, what's up with all the spacing changes?

In general, I think this looks good, but I'd like to review it more closely after you deal with anything related to the above.

humphd:
This also seems to have broken a ton of tests somehow, see https://travis-ci.org/filerjs/filer/builds/419287914?utm_source=github_status&utm_medium=notification.  Maybe rebasing will help.

dsych:
It seems to be failing [here](https://github.com/filerjs/filer/blob/develop/tests/lib/test-utils.js#L89). Not sure why, since I am doing the same thing in the example above...

dsych:
The culprit was parcel not being able to bundle node's util library. So instead I opted for an equivalent npm package.

dsych:
Another thing, one can argue that we don't need to write unit tests for any of the promise based api here, since they are re-using the same code that is already tested.

dsych:
@humphd I added a couple of things, do you mind taking a look?

modeswitch:
> Another thing, one can argue that we don't need to write unit tests for any of the promise based api here, since they are re-using the same code that is already tested.

That's an assumption that happens to be true right now. Unit tests ensure that your assumption continues to be true going forward. It's precisely why we *do* want unit tests.

src/filesystem/interface.js
```diff
@@ -262,94 +263,98 @@ function FileSystem(options, callback) {
...
-    var error = fs.queueOrRun(function() {
-      var context = fs.provider.openReadWriteContext();
-
-      // Fail early if the filesystem is in an error state (e.g.,
-      // provider failed to open.
-      if(FS_ERROR === fs.readyState) {
-        var err = new Errors.EFILESYSTEMERROR('filesystem unavailable, operation canceled');
-        return callback.call(fs, err);
-      }
+  [
```
humphd: Are these tabs?  Can we not do all these unnecessary whitespace changes so the history isn't affected for things unrelated to your fix?

src/filesystem/interface.js
```diff
@@ -262,94 +263,98 @@ function FileSystem(options, callback) {
...
+        // call signature, with complete() as the callback/last-arg now:
+        // fn(fs, context, arg0, arg1, ... , complete);
+        var fnArgs = [fs, context].concat(args);
+        impl[methodName].apply(null, fnArgs);
+      });
+      if(error) {
+        callback(error);
+      }
+    };
+    FileSystem.prototype.promises[methodName] = promisify(FileSystem.prototype[methodName].bind(fs));
```
humphd: Newline above this please.

package.json
```diff
@@ -35,6 +35,7 @@
   },
   "dependencies": {
     "base64-arraybuffer": "^0.1.5",
+    "es6-promisify": "^6.0.0",
```
humphd: Why this vs. the builtin `util.promisify`?

src/filesystem/interface.js
```diff
@@ -262,94 +263,98 @@ function FileSystem(options, callback) {
...
+  });
+  FileSystem.prototype.promises['exists'] = function() {
```
humphd: Newline above this, and also please add a comment block above this explaining why we special case this method.

src/filesystem/interface.js
dsych: This is done because I want to bind `this` reference to `promises` object in the same loop that regular functions are attached to the prototype. For that, I need to put the loop **inside** the constructor, this is where extra spaces are coming from. Previously the loop was in the global scope, now it is inside the function.

package.json
dsych: As I mentioned previously, for some reason parcel is not able to bundle `utils` module from node.

> The culprit was parcel not being able to bundle node's util library. So instead I opted for an equivalent npm package.

README.md
```diff
@@ -110,6 +110,21 @@ Like node.js, callbacks for methods that accept them are optional but suggested
 you omit the callback, errors will be thrown as exceptions). The first callback parameter is
 reserved for passing errors. It will be `null` if no errors occurred and should always be checked.
 
+#### Support for Promises is here!
```
humphd: Change this to "Support for Promises" please.

README.md
```diff
@@ -110,6 +110,21 @@ Like node.js, callbacks for methods that accept them are optional but suggested
...
+Promise based API mimics the way Node [implements](https://nodejs.org/api/fs.html#fs_fs_promises_api) them. Both `Shell` and `FileSystem` now have `promises` object attached alongside regular callback style methods. Method names are identical to their callback counterparts with the difference that instead of receiving third argument as a callback, they return a Promise that is resolved or rejected based on the success of method execution.
```
humphd: A few things to fix in this paragraph:

* "Promise based API" should be "The Promise based API"
* "now have `promises` object" should be "now have a `promises` object"
* "alongside regular callback style methods" should be "alongside the regular callback style methods"
* "instead of receiving third argument as a callback" should be "instead of taking a final callback argument"

Otherwise, looks great.

README.md
```diff
@@ -110,6 +110,21 @@ Like node.js, callbacks for methods that accept them are optional but suggested
...
+> Please note that `exists` method will always throw, since it was [deprecated](https://nodejs.org/api/fs.html#fs_fs_exists_path_callback) by Node.
+
+See example below:
+
+```javascript
+const fs = new Filer.FileSystem().promises;
+fs.open('/myfile', 'w+')
+  .then(fd => fs.close(fd))
+  .then(() => fs.stat('/myfile'))
```
humphd: Let's do two things here:

1. Let's also show the non-promises way of doing this, so people can compare the difference.  That is, let's have two code blocks and call out the difference between the two so people can clearly understand what you're doing here with respect to the original API.

2. Let's consider using `fstat` above, and closing the file descriptor toward the end, so the example is less contrived.  So the order would be open, fstat, close.

dsych: I used example from the Overview section to keep it consistent. I could just link my example to the Overview or re-write it completely as you suggested using `fstat.`

I am inclined to go with the first option of mere linking, since it will reduce clutter of an already code-rich readme. Let me know what you think.

dsych: We can still indicate the differences between styles while linking the examples.

README.md
```diff
@@ -110,6 +110,21 @@ Like node.js, callbacks for methods that accept them are optional but suggested
 you omit the callback, errors will be thrown as exceptions). The first callback parameter is
 reserved for passing errors. It will be `null` if no errors occurred and should always be checked.
 
+#### Support for Promises!
```
humphd: Remove the '!'.

src/filesystem/interface.js
```diff
@@ -262,94 +263,101 @@ function FileSystem(options, callback) {
...
-      // Forward this call to the impl's version, using the following
-      // call signature, with complete() as the callback/last-arg now:
-      // fn(fs, context, arg0, arg1, ... , complete);
-      var fnArgs = [fs, context].concat(args);
-      impl[methodName].apply(null, fnArgs);
-    });
-    if(error) {
-      callback(error);
-    }
+  // explicitly throw for 'exists' method, since it is deprecated and the equivalent promise-based implementation does not exist.
```
modeswitch: It's safe to omit this for the promise API. If someone tries to use it they will get an error anyway and it's less code to maintain.

dsych:
@modeswitch @humphd ok I think it is ready for a final review.

dsych:
> I'm happy for unit tests to come in follow-up PRs vs. needing it all here. But we will indeed want them for all methods. I'll set my students loose on the problem in a few weeks.

modeswitch:
@dsych OK, sounds good.

r+

humphd:
Great job @dsych, thank you for pushing this into the tree!

Issue #389, opened 2018-09-17
Add a test for truncate when length is undefined


humphd:
I noticed that there isn't a test for the case that we do:

```js
fs.truncate('file', undefined, function(err) {...});
```

The [docs for node.js](https://nodejs.org/api/fs.html#fs_fs_truncate_path_len_callback) say that this should assume `0`.

I would like to work on this if that's OK.

humphd:
Closing for now, so I can show my other section how this works. I'll create a new ticket and PR later.

PR #390, opened 2018-09-17
Fix Issue #389 - add test for truncate with undefined length

humphd:
This adds the test to make sure that we assume a `0` length if length is undefined.


humphd:
Closing for now, as I want to show this to my other students later today.  I'll do a new PR.

Issue #399, opened 2018-09-18
Adding testing with promises for fs.AppendFile()


JoshuaRM:
I noticed there are no existing test cases for the AppendFile method with promises, so I have included those. These new tests include the following specifications:

 fs.promises.appendFile:
-      should be a function
-      should append a utf8 file without specifying utf8 in appendFile
-      should append a utf8 file with "utf8" option to appendFile
-      should append a utf8 file with {encoding: "utf8"} option to appendFile
-      should append a binary file
-      should follow symbolic links
-      should work when file does not exist, and create the file

All of these tests have been successfully passed.

PR #401, opened 2018-09-18
Fix Issue #399 - added test cases for appendFile with promises

JoshuaRM:
New test cases have been added to the AppendFile tester, and updated to test using the fs.promises.appendFile method. This method now passes the following tests while using promises:

-      should be a function
-      should append a utf8 file without specifying utf8 in appendFile
-      should append a utf8 file with "utf8" option to appendFile
-      should append a utf8 file with {encoding: "utf8"} option to appendFile
-      should append a binary file
-      should follow symbolic links
-      should work when file does not exist, and create the file

All of these test are successfully passed while using promises

### DIFF
--- a/tests/spec/fs.appendFile.spec.js
+++ b/tests/spec/fs.appendFile.spec.js
@@ -125,3 +125,168 @@ describe('fs.appendFile', function() {
     });
   });
 });
+
+
+describe('fs.promises.appendFile', function() {
+  beforeEach(function(done) {
+    util.setup(function() {
+      var fs = util.fs();
+      fs.promises.writeFile('/myfile', 'This is a file.', { encoding: 'utf8' })
+        .then(done())
+        .catch(function (error)
+        {
+          throw error;
+        }); 
+    });
+  });
+  afterEach(util.cleanup);
+
+  it('should be a function', function() {
+    var fs = util.fs();
+    expect(fs.promises.appendFile).to.be.a('function');
+  });
+
+  it('should append a utf8 file without specifying utf8 in appendFile', function(done) {
+    var fs = util.fs();
+    var contents = 'This is a file.';
+    var more = ' Appended.';
+
+    fs.promises.appendFile('/myfile', more)
+      .then(function()
+      {
+        fs.readFile('/myfile', 'utf8', function(error, data) {
+          expect(error).not.to.exist;
+          expect(data).to.equal(contents + more);
+          done(); 
+        });
+      })
+      .catch(function(error)
+      {
+        throw error;
+      });
+  });
+
+  it('should append a utf8 file with "utf8" option to appendFile', function(done) {
+    var fs = util.fs();
+    var contents = 'This is a file.';
+    var more = ' Appended.';
+
+    fs.promises.appendFile('/myfile', more, 'utf8')
+      .then(function() {
+        fs.readFile('/myfile', 'utf8', function(error, data) {
+          expect(error).not.to.exist;
+          expect(data).to.equal(contents + more);
+          done();
+        });
+      })
+      .catch(function(error) {
+        throw error;
+      });
+  });
+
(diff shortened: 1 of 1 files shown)


jtj9817:
The planned changes detailed in issue #399 were implemented successfully and I ran your branch changes on my comp to see if there's going to be any issues but I've encountered none. Good job! Although one recommendation that I can suggest, having also had issues with UTF-8 encoding and decoding in another software, is that you should try to use Japanese/Korean/Chinese characters in the string for the method that you want to test and see if it still works as those set of characters is what the UTF-8 encoding is supposed to cover compared to other characters encoding.

tests/spec/fs.appendFile.spec.js
```diff
@@ -125,3 +125,168 @@ describe('fs.appendFile', function() {
     });
   });
 });
+
+
+describe('fs.promises.appendFile', function() {
+  beforeEach(function(done) {
+    util.setup(function() {
+      var fs = util.fs();
+      fs.promises.writeFile('/myfile', 'This is a file.', { encoding: 'utf8' })
```
humphd: Get rid of the `done` arg on line 131, and just return the promise like this:

```js
util.setup(() => util.fs().promises.writeFile('/myfile', 'This is a file'));
```

tests/spec/fs.appendFile.spec.js
```diff
@@ -125,3 +125,168 @@ describe('fs.appendFile', function() {
...
+  });
+
+  it('should append a utf8 file without specifying utf8 in appendFile', function(done) {
+    var fs = util.fs();
+    var contents = 'This is a file.';
+    var more = ' Appended.';
+
+    fs.promises.appendFile('/myfile', more)
+      .then(function()
+      {
```
humphd: Move your `{` to the same line as the function.

tests/spec/fs.appendFile.spec.js
```diff
@@ -125,3 +125,168 @@ describe('fs.appendFile', function() {
...
+        fs.readFile('/myfile', 'utf8', function(error, data) {
+          expect(error).not.to.exist;
+          expect(data).to.equal(contents + more);
+          done(); 
+        });
+      })
+      .catch(function(error)
```
humphd: You can get rid of this I think, just let the test deal with the failed case.

tests/spec/fs.appendFile.spec.js
```diff
@@ -125,3 +125,168 @@ describe('fs.appendFile', function() {
...
+    });
+  });
+  afterEach(util.cleanup);
+
+  it('should be a function', function() {
+    var fs = util.fs();
+    expect(fs.promises.appendFile).to.be.a('function');
+  });
+
+  it('should append a utf8 file without specifying utf8 in appendFile', function(done) {
```
humphd: Get rid of `done` here.

tests/spec/fs.appendFile.spec.js
```diff
@@ -125,3 +125,168 @@ describe('fs.appendFile', function() {
...
+    var fs = util.fs();
+    var contents = 'This is a file.';
+    var more = ' Appended.';
+
+    fs.promises.appendFile('/myfile', more)
```
humphd: Add a `return` to the front of this, so you return this promise (instead of needing/using `done`).

tests/spec/fs.appendFile.spec.js
```diff
@@ -125,3 +125,168 @@ describe('fs.appendFile', function() {
...
+      .then(function()
+      {
+        fs.readFile('/myfile', 'utf8', function(error, data) {
+          expect(error).not.to.exist;
+          expect(data).to.equal(contents + more);
+          done();
```
humphd: Get rid of this.

humphd:
Finished and landed in d1dd5fef6178320e4d889e95e06d5bd68c7ace83.  Make sure you finish what you start.

Issue #398, opened 2018-09-18
Add testing for promises to all test cases in fs.writeFile-readFile.spec.js file


alexander-ponomaroff:
Currently the fs.writeFile-readFile.spec.js file in the tests/spec folder tests the following cases for the fs.readFile() and fs.writeFile() methods:

    ✔ should be a function
    ✔ should error when path is wrong to readFile
    ✔ should write, read a utf8 file without specifying utf8 in writeFile
    ✔ should write, read a utf8 file with "utf8" option to writeFile
    ✔ should write, read a utf8 file with {encoding: "utf8"} option to writeFile
    ✔ should write, read a binary file
    ✔ should follow symbolic links

I will be performing the same tests with fs.promises.readFile() and fs.promises.writeFile() to ensure that the integration of promises in filer works as expected with readFile and writeFile methods.

In addition, I will also check that fs.promises.readFile() and fs.promises.writeFile() methods return a promise as expected ([specified in node.js docs](https://nodejs.org/api/fs.html#fs_fspromises_readfile_path_options)).

PR #402, opened 2018-09-18
Fix issue #398 - Add testing for promises to all test cases in fs.writeFile-readFile.spec.js file

alexander-ponomaroff:
I rewrote all test cases in the fs.writeFile-readFile.spec.js file to work with promises, testing that integration of promises into the readFile and writeFile methods works as expected.

In addition, I also check that fs.promises.readFile() and fs.promises.writeFile() methods return a promise as expected.

I added this code at the bottom of the file after the original test cases for readFile and writeFile methods.

### DIFF
--- a/tests/spec/fs.writeFile-readFile.spec.js
+++ b/tests/spec/fs.writeFile-readFile.spec.js
@@ -124,6 +124,7 @@ describe('fsPromises.writeFile, fsPromises.readFile', function() {
     p.then(() => {
       expect(fsPromises.readFile('/myfile', 'utf8')).to.be.a('Promise');
     });
+    return p;
   });
 
   it('should error when path is wrong to readFile', function() {
@@ -136,72 +137,52 @@ describe('fsPromises.writeFile, fsPromises.readFile', function() {
       });
   });
 
-  it('should write, read a utf8 file without specifying utf8 in writeFile', function(done) {
+  it('should write, read a utf8 file without specifying utf8 in writeFile', function() {
     var fsPromises = util.fs().promises;
     var contents = 'This is a file.';
 
-    fsPromises.writeFile('/myfile', contents)
+    return fsPromises.writeFile('/myfile', contents)
       .then( () => fsPromises.readFile('/myfile', 'utf8'))
-      .then(data => { 
-        expect(data).to.equal(contents);
-        done(); 
-      })
-      .catch(error => { throw error; });
+      .then(data => { expect(data).to.equal(contents); });
   });
 
-  it('should write, read a utf8 file with "utf8" option to writeFile', function(done) {
+  it('should write, read a utf8 file with "utf8" option to writeFile', function() {
     var fsPromises = util.fs().promises;
     var contents = 'This is a file.';
 
-    fsPromises.writeFile('/myfile', contents, 'utf8') 
+    return fsPromises.writeFile('/myfile', contents, 'utf8') 
       .then( () => fsPromises.readFile('/myfile', 'utf8'))
-      .then(data => { 
-        expect(data).to.equal(contents); 
-        done(); 
-      })  
-      .catch(error => { throw error; });
+      .then(data => { expect(data).to.equal(contents); });
   });
 
-  it('should write, read a utf8 file with {encoding: "utf8"} option to writeFile', function(done) {
+  it('should write, read a utf8 file with {encoding: "utf8"} option to writeFile', function() {
     var fsPromises = util.fs().promises;
     var contents = 'This is a file.';
 
-    fsPromises.writeFile('/myfile', contents, { encoding: 'utf8' })
+    return fsPromises.writeFile('/myfile', contents, { encoding: 'utf8' })
       .then( () => fsPromises.readFile('/myfile', 'utf8'))
-      .then(data => { 
-        expect(data).to.equal(contents);
-        done(); 
-      }) 
-      .catch(error => { throw error; });
+      .then(data => { expect(data).to.equal(contents); });
   });
 
-  it('should write, read a binary file', function(done) {
+  it('should write, read a binary file', function() {
     var fsPromises = util.fs().promises;
     // String and utf8 binary encoded versions of the same thing: 'This is a file.'
     var binary = new Buffer([84, 104, 105, 115, 32, 105, 115, 32, 97, 32, 102, 105, 108, 101, 46]);
 
-    fsPromises.writeFile('/myfile', binary)
+    return fsPromises.writeFile('/myfile', binary)
       .then( () => fsPromises.readFile('/myfile'))
-      .then(data => { 
-        expect(data).to.deep.equal(binary); 
-        done(); 
-      }) 
-      .catch(error => { throw error; });
+      .then(data => { expect(data).to.deep.equal(binary); });
   });
 
-  it('should follow symbolic links', function(done) {
+  it('should follow symbolic links', function() {
     var fsPromises = util.fs().promises;
     var contents = 'This is a file.';
 
-    fsPromises.writeFile('/myfile', '', { encoding: 'utf8' })
+    return fsPromises.writeFile('/myfile', '', { encoding: 'utf8' })
       .then( () => fsPromises.symlink('/myfile', '/myFileLink'))
       .then( () => fsPromises.writeFile('/myFileLink', contents, 'utf8'))
       .then( () => fsPromises.readFile('/myFileLink', 'utf8'))
-      .then(data => { 
-        expect(data).to.equal(contents); 
-        done(); 
-      }) 
-      .catch(error => { throw error; });
+      .then(data => { expect(data).to.equal(contents); });
   });
 });
 
--- a/tests/spec/fs.writeFile-readFile.spec.js
+++ b/tests/spec/fs.writeFile-readFile.spec.js
@@ -99,3 +99,118 @@ describe('fs.writeFile, fs.readFile', function() {
     });
   });
 });
+
+
+
+
+
+describe('fs.promises.writeFile, fs.promises.readFile', function() {
+  beforeEach(util.setup);
+  afterEach(util.cleanup);
+
+  it('should be a function', function() {
+    var fs = util.fs();
+    expect(fs.promises.writeFile).to.be.a('function');
+    expect(fs.promises.readFile).to.be.a('function');
+  });
+
+  it('should return a promise', function() {
+    var fs = util.fs();
+    var contents = 'This is a file.';
+
+    expect(fs.promises.writeFile('/myfile', contents)).to.be.a('Promise');
+    expect(fs.promises.readFile('/myfile', 'utf8')).to.be.a('Promise');
+  });
+
+  it('should error when path is wrong to readFile', function(done) {
+    var fs = util.fs();
+
+    
+    fs.promises.readFile('/no-such-file', 'utf8')
+      .then(data => expect(data).not.to.exist)
+      .catch(error => { expect(error).to.exist; expect(error.code).to.equal('ENOENT'); done(); });
+    
+
+  });
+
+  it('should write, read a utf8 file without specifying utf8 in writeFile', function(done) {
+    var fs = util.fs();
+    var contents = 'This is a file.';
+
+    fs.promises.writeFile('/myfile', contents)
+      .then( () => {
+
+        fs.promises.readFile('/myfile', 'utf8')
+          .then( (data, error) => { expect(data).to.equal(contents); expect(error).not.to.exist; done(); });
+
+      })
+      .catch(error => { throw error; });
+  });
+
+  it('should write, read a utf8 file with "utf8" option to writeFile', function(done) {
+    var fs = util.fs();
+    var contents = 'This is a file.';
+
+    fs.promises.writeFile('/myfile', contents, 'utf8') 
+      .then( () => {
+
+        fs.promises.readFile('/myfile', 'utf8')
+          .then( (data, error) => { expect(data).to.equal(contents); expect(error).not.to.exist; done(); });
+
+      })  
+      .catch(error => { throw error; });
+  });
+
+  it('should write, read a utf8 file with {encoding: "utf8"} option to writeFile', function(done) {
+    var fs = util.fs();
+    var contents = 'This is a file.';
+
+    fs.promises.writeFile('/myfile', contents, { encoding: 'utf8' })
+      .then( () => {
+
+        fs.promises.readFile('/myfile', 'utf8')
+          .then( (data, error) => { expect(data).to.equal(contents); expect(error).not.to.exist; done(); });
+
+      }) 
+      .catch(error => { throw error; });
+  });
+
+  it('should write, read a binary file', function(done) {
+    var fs = util.fs();
+    // String and utf8 binary encoded versions of the same thing: 'This is a file.'
+    var binary = new Buffer([84, 104, 105, 115, 32, 105, 115, 32, 97, 32, 102, 105, 108, 101, 46]);
+
+    fs.promises.writeFile('/myfile', binary)
+      .then( () => {
+
+        fs.promises.readFile('/myfile')
+          .then( (data, error) => { expect(data).to.deep.equal(binary); expect(error).not.to.exist; done(); });
+
+      }) 
+      .catch(error => { throw error; });
+  });
+
+  it('should follow symbolic links', function(done) {
+    var fs = util.fs();
+    var contents = 'This is a file.';
+
+    fs.promises.writeFile('/myfile', '', { encoding: 'utf8' })
+      .then( () => {
+        fs.promises.symlink('/myfile', '/myFileLink')
+          .then( () => {
+            fs.promises.writeFile('/myFileLink', contents, 'utf8')
+              .then( () => {
+
+                fs.promises.readFile('/myFileLink', 'utf8')
+                  .then( (data, error) => { expect(data).to.equal(contents); expect(error).not.to.exist; done(); });
+                  
+              }) 
+              .catch(error => { throw error; });   
+          }) 
+          .catch(error => { throw error; });
+      }) 
+      .catch(error => { throw error; });
+  });
+});
+
+
--- a/tests/spec/fs.writeFile-readFile.spec.js
+++ b/tests/spec/fs.writeFile-readFile.spec.js
@@ -100,114 +100,128 @@ describe('fs.writeFile, fs.readFile', function() {
   });
 });
 
+/**
+ * fsPromises tests
+ */
 
-
-
-
-describe('fs.promises.writeFile, fs.promises.readFile', function() {
+describe('fsPromises.writeFile, fsPromises.readFile', function() {
   beforeEach(util.setup);
   afterEach(util.cleanup);
 
   it('should be a function', function() {
-    var fs = util.fs();
-    expect(fs.promises.writeFile).to.be.a('function');
-    expect(fs.promises.readFile).to.be.a('function');
+    var fsPromises = util.fs().promises;
+    expect(fsPromises.writeFile).to.be.a('function');
+    expect(fsPromises.readFile).to.be.a('function');
   });
 
   it('should return a promise', function() {
-    var fs = util.fs();
+    var fsPromises = util.fs().promises;
     var contents = 'This is a file.';
 
-    expect(fs.promises.writeFile('/myfile', contents)).to.be.a('Promise');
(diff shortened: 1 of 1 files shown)


tests/spec/fs.writeFile-readFile.spec.js
```diff
@@ -99,3 +99,118 @@ describe('fs.writeFile, fs.readFile', function() {
     });
   });
 });
+
+
+
+
+
```
humphd: I think you're trying to separate the callback from promise based tests with these blank lines.  Could you instead do the following:

```js
/**
 * fsPromises tests
 */
...
```

tests/spec/fs.writeFile-readFile.spec.js
```diff
@@ -99,3 +99,118 @@ describe('fs.writeFile, fs.readFile', function() {
...
+describe('fs.promises.writeFile, fs.promises.readFile', function() {
+  beforeEach(util.setup);
+  afterEach(util.cleanup);
+
+  it('should be a function', function() {
+    var fs = util.fs();
```
humphd: This will work, but it's more common for people to grab the `promises` object and cache it here, like so:

```js
...
fsPromises.writeFile()...
```

This way you don't have to do `fs.promises` all the time below.  This same change could happen throughout your code.

tests/spec/fs.writeFile-readFile.spec.js
```diff
@@ -99,3 +99,118 @@ describe('fs.writeFile, fs.readFile', function() {
...
+    var contents = 'This is a file.';
+
+    expect(fs.promises.writeFile('/myfile', contents)).to.be.a('Promise');
+    expect(fs.promises.readFile('/myfile', 'utf8')).to.be.a('Promise');
+  });
+
+  it('should error when path is wrong to readFile', function(done) {
+    var fs = util.fs();
+
+
```
humphd: Just one blank line here please.

tests/spec/fs.writeFile-readFile.spec.js
```diff
@@ -99,3 +99,118 @@ describe('fs.writeFile, fs.readFile', function() {
...
+  });
+
+  it('should error when path is wrong to readFile', function(done) {
+    var fs = util.fs();
+
+    
+    fs.promises.readFile('/no-such-file', 'utf8')
+      .then(data => expect(data).not.to.exist)
+      .catch(error => { expect(error).to.exist; expect(error.code).to.equal('ENOENT'); done(); });
+
```
humphd: These blank lines can be removed.

tests/spec/fs.writeFile-readFile.spec.js
```diff
@@ -99,3 +99,118 @@ describe('fs.writeFile, fs.readFile', function() {
...
+    expect(fs.promises.readFile('/myfile', 'utf8')).to.be.a('Promise');
+  });
+
+  it('should error when path is wrong to readFile', function(done) {
+    var fs = util.fs();
+
+    
+    fs.promises.readFile('/no-such-file', 'utf8')
+      .then(data => expect(data).not.to.exist)
+      .catch(error => { expect(error).to.exist; expect(error.code).to.equal('ENOENT'); done(); });
```
humphd: It's always better to split your lines up vs. using long lines like this, because it's easy to miss things when you're reading the code.

tests/spec/fs.writeFile-readFile.spec.js
```diff
@@ -99,3 +99,118 @@ describe('fs.writeFile, fs.readFile', function() {
...
+
+  it('should write, read a utf8 file without specifying utf8 in writeFile', function(done) {
+    var fs = util.fs();
+    var contents = 'This is a file.';
+
+    fs.promises.writeFile('/myfile', contents)
+      .then( () => {
+
+        fs.promises.readFile('/myfile', 'utf8')
+          .then( (data, error) => { expect(data).to.equal(contents); expect(error).not.to.exist; done(); });
```
humphd: Break this up into multiple lines.  Also, the `error` object won't ever get passed here, it will end up in the `catch`

tests/spec/fs.writeFile-readFile.spec.js
```diff
@@ -99,3 +99,118 @@ describe('fs.writeFile, fs.readFile', function() {
...
+
+  it('should write, read a utf8 file with "utf8" option to writeFile', function(done) {
+    var fs = util.fs();
+    var contents = 'This is a file.';
+
+    fs.promises.writeFile('/myfile', contents, 'utf8') 
+      .then( () => {
+
+        fs.promises.readFile('/myfile', 'utf8')
+          .then( (data, error) => { expect(data).to.equal(contents); expect(error).not.to.exist; done(); });
```
humphd: Same thing here, break it up, dump the `error`

tests/spec/fs.writeFile-readFile.spec.js
```diff
@@ -99,3 +99,118 @@ describe('fs.writeFile, fs.readFile', function() {
...
+
+  it('should write, read a utf8 file with {encoding: "utf8"} option to writeFile', function(done) {
+    var fs = util.fs();
+    var contents = 'This is a file.';
+
+    fs.promises.writeFile('/myfile', contents, { encoding: 'utf8' })
+      .then( () => {
+
+        fs.promises.readFile('/myfile', 'utf8')
+          .then( (data, error) => { expect(data).to.equal(contents); expect(error).not.to.exist; done(); });
```
humphd: And here.

tests/spec/fs.writeFile-readFile.spec.js
```diff
@@ -99,3 +99,118 @@ describe('fs.writeFile, fs.readFile', function() {
...
+  it('should write, read a binary file', function(done) {
+    var fs = util.fs();
+    // String and utf8 binary encoded versions of the same thing: 'This is a file.'
+    var binary = new Buffer([84, 104, 105, 115, 32, 105, 115, 32, 97, 32, 102, 105, 108, 101, 46]);
+
+    fs.promises.writeFile('/myfile', binary)
+      .then( () => {
+
+        fs.promises.readFile('/myfile')
+          .then( (data, error) => { expect(data).to.deep.equal(binary); expect(error).not.to.exist; done(); });
```
humphd: And here.

tests/spec/fs.writeFile-readFile.spec.js
```diff
@@ -99,3 +99,118 @@ describe('fs.writeFile, fs.readFile', function() {
...
+
+                fs.promises.readFile('/myFileLink', 'utf8')
+                  .then( (data, error) => { expect(data).to.equal(contents); expect(error).not.to.exist; done(); });
+                  
+              }) 
+              .catch(error => { throw error; });   
+          }) 
+          .catch(error => { throw error; });
+      }) 
+      .catch(error => { throw error; });
```
humphd: Instead of having 3 of these, I think we can just have one final `catch` on the outer promise chain.

tests/spec/fs.writeFile-readFile.spec.js
```diff
@@ -99,3 +99,118 @@ describe('fs.writeFile, fs.readFile', function() {
...
+
+    fs.promises.writeFile('/myfile', '', { encoding: 'utf8' })
+      .then( () => {
+        fs.promises.symlink('/myfile', '/myFileLink')
+          .then( () => {
+            fs.promises.writeFile('/myFileLink', contents, 'utf8')
+              .then( () => {
+
+                fs.promises.readFile('/myFileLink', 'utf8')
+                  .then( (data, error) => { expect(data).to.equal(contents); expect(error).not.to.exist; done(); });
```
humphd: And here.

tests/spec/fs.writeFile-readFile.spec.js
alexander-ponomaroff: What do you mean by dump the error?
I use error to check this: expect(error).not.to.exist;

alexander-ponomaroff: Nevermind, for some reason the comment above where you explain this was hidden. I just saw it.

tests/spec/fs.writeFile-readFile.spec.js
alexander-ponomaroff: Thank you very much for the feedback, I will change all of these soon.

jespirit:
You should consider rewriting the tests to look more like the following:

```javascript
it('should write, read a utf8 file with {encoding: "utf8"} option to writeFile', function(done) {
  var contents = 'This is a file.';
  
  fsPromises.writeFile('/myfile', contents, { encoding: 'utf8' })
    .then(fsPromises.readFile('/myfile', 'utf8'))  // here you can call another function directly inside a then() handler
    .then(data => {
      expect(data).to.equal(contents);
      done();
    })
    .catch(error => { throw error; });  // just have only one catch handler at the very end
                                        // will catch any error that occur anywhere along the chain
});
```

Chaining promises make it so that you can order the functions in sequence, to be called one after another, and then have a catch handler at the very end. This approach is modeled after how synchronous code works. This is a really good guide on promises: [Using Promises](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises).

```javascript
try {  
  fs.writeFile('/myfile', contents, { encoding: 'utf8' });
  data = fs.readFile('/myfile', 'utf8');
  if (data !== contents) throw error;
} catch (error) {  // catch any and all errors, whether they occured on the 1st, 2nd, or 3rd line
  failureCallback(error);
}
```

tests/spec/fs.writeFile-readFile.spec.js
```diff
@@ -99,3 +99,132 @@ describe('fs.writeFile, fs.readFile', function() {
...
+  });
+
+  it('should write, read a utf8 file with "utf8" option to writeFile', function(done) {
+    var fsPromises = util.fs().promises;
+    var contents = 'This is a file.';
+
+    fsPromises.writeFile('/myfile', contents, 'utf8') 
+      .then( () => {
+
+        fsPromises.readFile('/myfile', 'utf8')
```
jespirit: Same as before, move this function call inside the `then()` handler.
eg. `then(fsPromises.readFile('myfile', 'utf8')`

tests/spec/fs.writeFile-readFile.spec.js
```diff
@@ -99,3 +99,132 @@ describe('fs.writeFile, fs.readFile', function() {
...
+          .then(data => { 
+            expect(data).to.equal(contents); 
+            done(); 
+          })
+          .catch(error => expect(error).not.to.exist);
```
jespirit: Remove this

tests/spec/fs.writeFile-readFile.spec.js
```diff
@@ -99,3 +99,132 @@ describe('fs.writeFile, fs.readFile', function() {
...
+
+  it('should write, read a binary file', function(done) {
+    var fsPromises = util.fs().promises;
+    // String and utf8 binary encoded versions of the same thing: 'This is a file.'
+    var binary = new Buffer([84, 104, 105, 115, 32, 105, 115, 32, 97, 32, 102, 105, 108, 101, 46]);
+
+    fsPromises.writeFile('/myfile', binary)
+      .then( () => {
+
+        fsPromises.readFile('/myfile')
```
jespirit: Same as before, move this function call inside the `then()` handler.
eg. `fsPromises.readFile('/myfile')`

tests/spec/fs.writeFile-readFile.spec.js
```diff
@@ -99,3 +99,132 @@ describe('fs.writeFile, fs.readFile', function() {
...
+  });
+
+  it('should write, read a utf8 file with {encoding: "utf8"} option to writeFile', function(done) {
+    var fsPromises = util.fs().promises;
+    var contents = 'This is a file.';
+
+    fsPromises.writeFile('/myfile', contents, { encoding: 'utf8' })
+      .then( () => {
+
+        fsPromises.readFile('/myfile', 'utf8')
```
jespirit: Same as before, move this function call inside the `then()` handler.
eg. `fsPromises.readFile('/myfile', 'utf8')`

tests/spec/fs.writeFile-readFile.spec.js
```diff
@@ -99,3 +99,132 @@ describe('fs.writeFile, fs.readFile', function() {
...
+          .then(data => { 
+            expect(data).to.equal(contents);
+            done(); 
+          })
+          .catch(error => expect(error).not.to.exist);
```
jespirit: Remove this

tests/spec/fs.writeFile-readFile.spec.js
```diff
@@ -99,3 +99,132 @@ describe('fs.writeFile, fs.readFile', function() {
...
+
+    fsPromises.writeFile('/myfile', binary)
+      .then( () => {
+
+        fsPromises.readFile('/myfile')
+          .then(data => { 
+            expect(data).to.deep.equal(binary); 
+            done(); 
+          })
+          .catch(error => expect(error).not.to.exist);
```
jespirit: Remove this

tests/spec/fs.writeFile-readFile.spec.js
```diff
@@ -99,3 +99,132 @@ describe('fs.writeFile, fs.readFile', function() {
...
+      .catch(error => { throw error; });
+  });
+
+  it('should follow symbolic links', function(done) {
+    var fsPromises = util.fs().promises;
+    var contents = 'This is a file.';
+
+    fsPromises.writeFile('/myfile', '', { encoding: 'utf8' })
+      .then( () => {
+        fsPromises.symlink('/myfile', '/myFileLink')
```
jespirit: Same as before, move this function call inside the `then()` handler.
eg. `fsPromises.symlink('/myfile', '/myFileLink')`

tests/spec/fs.writeFile-readFile.spec.js
```diff
@@ -99,3 +99,132 @@ describe('fs.writeFile, fs.readFile', function() {
...
+          .then( () => {
+            fsPromises.writeFile('/myFileLink', contents, 'utf8')
```
jespirit: Same as before, move this function call inside the `then()` handler.
eg. `fsPromises.writeFile('/myFileLink', contents, 'utf8')`

tests/spec/fs.writeFile-readFile.spec.js
```diff
@@ -99,3 +99,132 @@ describe('fs.writeFile, fs.readFile', function() {
...
+              .then( () => {
+
+                fsPromises.readFile('/myFileLink', 'utf8')
```
jespirit: Same as before, move this function call inside the `then()` handler.
eg. `fsPromises.readFile('/myFileLink', 'utf8')`

tests/spec/fs.writeFile-readFile.spec.js
```diff
@@ -99,3 +99,132 @@ describe('fs.writeFile, fs.readFile', function() {
...
+                  .then(data => { 
+                    expect(data).to.equal(contents); 
+                    done(); 
+                  })
+                  .catch(error => expect(error).not.to.exist);
```
jespirit: Remove this

jespirit:
Okay, to support my case I decided to convert Dave's truncate test https://github.com/filerjs/filer/pull/395 to use `fs.promises` using nested `then()` handlers.

```javascript
it('should assume a length of 0 if passed undefined', function(done) {
  var buffer = new Buffer([1, 2, 3, 4, 5, 6, 7, 8]);

  fsPromises.open('/myfile', 'w')
    .then(fd => {
      fsPromises.write(fd, buffer, 0, buffer.length, 0)
        .then(result => {
          // check intermediate results here
          expect(result).to.equal(buffer.length);
          fsPromises.close(fd)
            .then(() => {
              // We want to use undefined to see that it defaults to 0
              fsPromises.truncate('/myfile', undefined)
                .then(() => {
                  fsPromises.stat('/myfile')
                    .then(result => {
                      // check final results here
                      expect(result.size).to.equal(0);
                      done();
                    })
                  .catch(error => { throw error; });
                })
                .catch(error => { throw error; });
            })
            .catch(error => { throw error; });
        })
        .catch(error => { throw error; });
      })
      .catch(error => { throw error; });
});
```

I rewrote the above code without using nested `then()` handlers so that it looks more like synchronous code. It's much easier to read and you'll notice there's only a single `catch()` handler at the very end. No need to include so many `catch(error =>{ ... })` handlers.

```javascript
it('should assume a length of 0 if passed undefined', function(done) {
  var buffer = new Buffer([1, 2, 3, 4, 5, 6, 7, 8]);

  fsPromises.open('/myfile', 'w')
    .then(fd => fsPromises.write(fd, buffer, 0, buffer.length, 0))
    .then(result => {
      // check intermediate results here
      expect(result).to.equal(buffer.length);
      fsPromises.close(fd);
    })
    .then(() => {
      // We want to use undefined to see that it defaults to 0
      fsPromises.truncate('/myfile', undefined);
    })
    .then(() => fsPromises.stat('/myfile'))
    .then(result => {
      // check final results here
      expect(result.size).to.equal(0);
      done();
    })
    // catch any and all errors that occur anywhere along the chain
    .catch(error => { throw error; });
});
```

If you're wondering about the syntax, for example of something like in the following code snippet, it's important to know that `then` returns a `Promise` and its status (pending, resolved, or rejected) depends on the `onFulfilled` handler (eg. anonymous function `fd => {...}`). Because `fsPromises.write` returns a promise, that `Promise` object will be passed to the next `then` handler (or `catch` handler if an error occurred).

```javascript
.then(fd => fsPromises.write(fd, buffer, 0, buffer.length, 0))
// same as
.then(fd => return fsPromises.write(fd, buffer, 0, buffer.length, 0))
// same as
.then(function(fd) { return fsPromises.write(fd, buffer, 0, buffer.length, 0); })
```

[Link to MDN Promise.then()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then)

tests/spec/fs.writeFile-readFile.spec.js
```diff
@@ -99,3 +99,132 @@ describe('fs.writeFile, fs.readFile', function() {
...
+    expect(fsPromises.writeFile).to.be.a('function');
+    expect(fsPromises.readFile).to.be.a('function');
+  });
+
+  it('should return a promise', function() {
+    var fsPromises = util.fs().promises;
+    var contents = 'This is a file.';
+
+    expect(fsPromises.writeFile('/myfile', contents)).to.be.a('Promise');
+    expect(fsPromises.readFile('/myfile', 'utf8')).to.be.a('Promise');
```
humphd: There's a potential timing bug here.  Your second call to `readFile` *might* fail if the first `writeFile` hasn't finished yet.  You could deal with it like so:

```js
var p = fsPromises.writeFile('/myfile', contents);
expect(p).to.be.a('Promise');

p.then(() => {
  fsPromises.readFile('/myfile', 'utf8')).to.be.a('Promise');
});
```

tests/spec/fs.writeFile-readFile.spec.js
```diff
@@ -99,3 +99,132 @@ describe('fs.writeFile, fs.readFile', function() {
...
+  });
+
+  it('should error when path is wrong to readFile', function(done) {
+    var fsPromises = util.fs().promises;
+
+    fsPromises.readFile('/no-such-file', 'utf8')
+      .then(data => expect(data).not.to.exist)
```
humphd: I don't think you need the `.then()` at all, since you just want to deal with the `.catch()`.

tests/spec/fs.writeFile-readFile.spec.js
```diff
@@ -99,3 +99,132 @@ describe('fs.writeFile, fs.readFile', function() {
...
+
+  it('should return a promise', function() {
+    var fsPromises = util.fs().promises;
+    var contents = 'This is a file.';
+
+    expect(fsPromises.writeFile('/myfile', contents)).to.be.a('Promise');
+    expect(fsPromises.readFile('/myfile', 'utf8')).to.be.a('Promise');
+  });
+
+  it('should error when path is wrong to readFile', function(done) {
```
humphd: Get rid of the `done` callback here, and instead you need to add a `return` to the front of line 128 below, so you return your Promise.

alexander-ponomaroff:
@humphd I will fix these issues by Sunday. Am currently busy, but will have some time on the weekend to get everything in order.

humphd:
@alexander-ponomaroff how is this coming?

alexander-ponomaroff:
@humphd Really sorry about the delay, I just fixed all the issues that were requested. Thank you @jespirit for the suggestions, the code looks more neat now.

Issue #400, opened 2018-09-18
Add tests for fs.promises.readdir


jespirit:
I can rewrite the tests in [fs.readdir.spec.js](https://github.com/filerjs/filer/blob/master/tests/spec/fs.readdir.spec.js) to use Filer's Promises API

humphd:
There are a few tests possible for this.  For example, if the path you specify is not found, if the path is actually a file, if the path exists and is a dir, etc.  Just pick one and do that, and let others add the remaining cases.

jespirit:
I'll test `fs.promises.readdir` by passing in a file.

PR #406, opened 2018-09-19
Issue 396 - Add a test for fs.promises.stat() when the path does not exist

pynnl:
Fix #396 
<details><summary>Test log</summary>
<p>

```
START:
18 09 2018 20:47:16.270:INFO [karma]: Karma v3.0.0 server started at http://0.0.0.0:9876/
18 09 2018 20:47:16.272:INFO [launcher]: Launching browser ChromeHeadless with unlimited concurrency
18 09 2018 20:47:16.282:INFO [launcher]: Starting browser ChromeHeadless
18 09 2018 20:47:17.093:INFO [HeadlessChrome 0.0.0 (Windows 10 0.0.0)]: Connected on socket M2LWgwfiA9J7Fd8TAAAA with id 8357030
  fs.stat
    √ should follow symbolic links and return a stat object for the resulting path
    √ should return a stat object for a valid descriptor
  fsPromises.stat
    √ should return an error if path does not exist
  fs.lstat
    √ should be a function
    √ should return an error if path does not exist
    √ should return a stat object if path is not a symbolic link
    √ should return a stat object if path is a symbolic link
  fs.exists
    √ should be a function
WARN: 'This method is deprecated. For more details see https://nodejs.org/api/fs.html#fs_fs_exists_path_callback'
    √ should return false if path does not exist
WARN: 'This method is deprecated. For more details see https://nodejs.org/api/fs.html#fs_fs_exists_path_callback'
    √ should return true if path exists
WARN: 'This method is deprecated. For more details see https://nodejs.org/api/fs.html#fs_fs_exists_path_callback'
    √ should follow symbolic links and return true for the resulting path
  fs.mknod
    √ should be a function
    √ should return an error if part of the parent path does not exist
    √ should return an error if path already exists
    √ should return an error if the parent node is not a directory
    √ should return an error if the mode provided is not DIRECTORY or FILE
    √ should make a new directory
    √ should make a new file
  fs.mkdir
    √ should be a function
    √ should return an error if part of the parent path does not exist
    √ should return an error if the path already exists
    √ should make a new directory
  fs.readdir
    √ should be a function
    √ should return an error if the path does not exist
    √ should return a list of files from an existing directory
    √ should follow symbolic links
  fs.rmdir
    √ should be a function
    √ should return an error if the path does not exist
    √ should return an error if attempting to remove the root directory
    √ should return an error if the directory is not empty
    √ should return an error if the path is not a directory
    √ should return an error if the path is a symbolic link
    √ should remove an existing directory
[...]

PR #409, opened 2018-09-19
Fix Issue #400 - add test for passing file to fs.promise.readdir()

jespirit:
This adds a test to ensure that `fs.promise.readdir()` throws an error when passing in a file.

### DIFF
--- a/tests/spec/fs.readdir.spec.js
+++ b/tests/spec/fs.readdir.spec.js
@@ -55,3 +55,25 @@ describe('fs.readdir', function() {
     });
   });
 });
+
+/**
+ * fsPromises tests
+ */
+
+describe('fsPromises.readdir', function() {
+  beforeEach(util.setup);
+  afterEach(util.cleanup);
+
+  it('should return an error if the path is a file', function(done) {
+    var fsPromises = util.fs().promises;
+
+    fsPromises.open('/myfile', 'w')
+      .then(fd => fsPromises.close(fd))
+      .then(() => fsPromises.readdir('/myfile'))
+      .catch(error => {
+        expect(error).to.exist;
+        expect(error.code).to.equal('ENOTDIR');
+        done();
+      });
+  });
+});
--- a/tests/spec/fs.readdir.spec.js
+++ b/tests/spec/fs.readdir.spec.js
@@ -64,16 +64,15 @@ describe('fsPromises.readdir', function() {
   beforeEach(util.setup);
   afterEach(util.cleanup);
 
-  it('should return an error if the path is a file', function(done) {
+  it('should return an error if the path is a file', function() {
     var fsPromises = util.fs().promises;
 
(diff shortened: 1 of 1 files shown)


tests/spec/fs.readdir.spec.js
```diff
@@ -55,3 +55,25 @@ describe('fs.readdir', function() {
...
+
+/**
+ * fsPromises tests
+ */
+
+describe('fsPromises.readdir', function() {
+  beforeEach(util.setup);
+  afterEach(util.cleanup);
+
+  it('should return an error if the path is a file', function(done) {
```
rdittrich97: To make this more promise-friendly, please remove the `done` from the callback function argument

tests/spec/fs.readdir.spec.js
```diff
@@ -55,3 +55,25 @@ describe('fs.readdir', function() {
...
+    var fsPromises = util.fs().promises;
+
+    fsPromises.open('/myfile', 'w')
```
rdittrich97: To make this more promise-friendly, please return the return of this function.

tests/spec/fs.readdir.spec.js
```diff
@@ -55,3 +55,25 @@ describe('fs.readdir', function() {
...
+      .then(fd => fsPromises.close(fd))
+      .then(() => fsPromises.readdir('/myfile'))
+      .catch(error => {
+        expect(error).to.exist;
+        expect(error.code).to.equal('ENOTDIR');
+        done();
```
rdittrich97: To make this more promise-friendly, please remove the unnecessary  `done();`.

tests/spec/fs.readdir.spec.js
```diff
@@ -55,3 +55,25 @@ describe('fs.readdir', function() {
     });
   });
 });
+
+/**
+ * fsPromises tests
+ */
+
+describe('fsPromises.readdir', function() {
```
y2s82: Move the following test into the `describe('fs.readdir'...)` block.  This makes it easier to add more related tests if others are working on the same function. You may also need to change the description of each test to separate them from the original test. e.g. prefix them with `(promises)`

jespirit:
@rdittrich97 Thanks for the input. I've made the changes.

tests/spec/fs.readdir.spec.js
jespirit: I just think it would be better organized to separate the function callback and promises-based tests. I've looked over other pull requests and they seem to follow the same pattern. For example, `fs.promises.appendFile`: https://github.com/filerjs/filer/pull/401/commits/24b9dc7b457feb1dd4a87aa54ed51a47b143bbad and `fs.promises.stat`: https://github.com/filerjs/filer/pull/406/commits/d07080c75c1db06db20ba4263a1bc147ba685246

PR #412, opened 2018-09-19
Fix #387 - Add test to write data to a file using the Promise API

### LINKED ISSUE #387: Testing open, close and stat using the promise based API
```javascript
const fs = new Filer.FileSystem().promises;
fs.open('/myfile', 'w+')
  .then(fd => fs.close(fd))
  .then(() => fs.stat('/myfile'))
  .then(stats => { console.log(`stats: ${JSON.stringify(stats)}`); })
  .catch(err => { console.error(err); })
```
I'd like to write a test for the above example in the readme file. The test would check if the file was opened, closed, and whether the expected file information was returned.

dnguneratne:
This PR adds a new test to ensure that,

- a file was opened,
- a buffer of 8 numbers were successfully written to it,
- and check if the file is of type FILE,

using the Promise API.

### DIFF
--- a/tests/spec/fs.write.spec.js
+++ b/tests/spec/fs.write.spec.js
@@ -59,4 +59,31 @@ describe('fs.write', function() {
       });
     });
   });
+
+  /**
+   * fsPromise tests
+   */
+
+  it('should write data to a file using the Promise API', () => {
+    var fs = util.fs().promises;
+    var buffer = new Filer.Buffer([1, 2, 3, 4, 5, 6, 7, 8]);
+
+    return fs.open('/myfile', 'w+')
+      .then((fd) => {
+        return fs.write(fd, buffer, 0, buffer.length, 0)
+          .then((result) => {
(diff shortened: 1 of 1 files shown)


tests/spec/fs.write.spec.js
```diff
@@ -59,4 +59,31 @@ describe('fs.write', function() {
...
+      })
+      .then(() => { 
+        return fs.stat('/myfile')
+          .then((result) => {
+            expect(result.type).to.equal('FILE');
+            expect(result.size).to.equal(buffer.length);
+          });
+      })
+      .catch((error) => {
+        if(error) throw error;
```
klymenkoo: You don't need to check ```if(error)``` inside ```catch``` block. You can just throw error.

humphd:
I just landed a changed to update our promises api to match node's, and `.write()` doesn't exist anymore.  They put it on a `FileHandle` object instead.  Sorry about this, but the code is wrong now (not your fault, ours).  Closing.

PR #413, opened 2018-09-19
Fix Issue 410 add test for fs.rename() when file or directory to rename doesn't exist

itchylol742:
This adds a test for fs.rename() which expects an error when trying to rename a file/directory that doesn't exist.

```js
it('should return error if the file or directory to rename doesnt exist', function(done) {
    var fs = util.fs();
      fs.rename('/fakeDirectory', '/myotherdir', function(error) {
        expect(error).to.exist;
        done();
      });
  });
```

### DIFF
--- a/tests/spec/fs.rename.spec.js
+++ b/tests/spec/fs.rename.spec.js
@@ -69,6 +69,16 @@ describe('fs.rename', function() {
       });
     });
   });
+  //337
+  it('should error if the file or directory to rename doesnt exist', function(done) {
+    var fs = util.fs();
+
+      fs.rename('/fakeDirectory', '/myotherdir', function(error) {
+        expect(error).to.exist;
+        done();
+        
+      });
+  });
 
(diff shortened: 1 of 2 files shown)


tests/spec/fs.rename.spec.js
```diff
@@ -69,6 +69,16 @@ describe('fs.rename', function() {
       });
     });
   });
+  //337
+  it('should error if the file or directory to rename doesnt exist', function(done) {
+    var fs = util.fs();
+
+      fs.rename('/fakeDirectory', '/myotherdir', function(error) {
+        expect(error).to.exist;
+        done();
```
giantpanpan: Great Job! @itchylol742 
I just want to add one my opinion that if you want to test the existing of a path/file,
it might be a good way to add:

Thank you :)

humphd:
@itchylol742 This can't be merged in its current form: you've got changes to unrelated files, and tons of whitespace changes.  I'm going to close this, and if you want to file a new PR with just your test changes, that would be great.

Issue #407, opened 2018-09-19
Add tests for fsPromises.mkdir


klymenkoo:
I found that there are no tests for fs.promises version of mkdir method. 
The new tests can be identical to [fs.mkdir version](https://github.com/filerjs/filer/blob/master/tests/spec/fs.mkdir.spec.js):

```
should be a function 
```
```
should return an error if part of the parent path does not exist 
```
```
should return an error if the path already exists 
```
```
should make a new directory
```

I would also include tests to check the return type of fsPromises.mkdir:
```
should return Promise
```

I would like to take this issue and work on it if it is possible.

PR #415, opened 2018-09-20
Fix issue #407 - Add tests for Promises version of mkdir

vldmrkl:
Fix for issue #407 
This PR contains tests for ```fs.promises.mkdir```  : 

- should be a function
- should return an error if part of the parent path does not exist
- should return an error if the path already exists
- should make a new directory
- should return a promise

### DIFF
--- a/tests/spec/fs.mkdir.spec.js
+++ b/tests/spec/fs.mkdir.spec.js
@@ -1,43 +1,43 @@
 var util = require('../lib/test-utils.js');
 var expect = require('chai').expect;
 
-describe('fs.mkdir', function() {
+describe('fs.mkdir', function () {
   beforeEach(util.setup);
   afterEach(util.cleanup);
 
-  it('should be a function', function() {
+  it('should be a function', function () {
     var fs = util.fs();
     expect(fs.mkdir).to.be.a('function');
   });
 
-  it('should return an error if part of the parent path does not exist', function(done) {
+  it('should return an error if part of the parent path does not exist', function (done) {
     var fs = util.fs();
 
-    fs.mkdir('/tmp/mydir', function(error) {
+    fs.mkdir('/tmp/mydir', function (error) {
       expect(error).to.exist;
       expect(error.code).to.equal('ENOENT');
       done();
     });
   });
 
-  it('should return an error if the path already exists', function(done) {
+  it('should return an error if the path already exists', function (done) {
     var fs = util.fs();
 
-    fs.mkdir('/', function(error) {
+    fs.mkdir('/', function (error) {
       expect(error).to.exist;
       expect(error.code).to.equal('EEXIST');
       done();
     });
   });
 
-  it('should make a new directory', function(done) {
+  it('should make a new directory', function (done) {
     var fs = util.fs();
 
-    fs.mkdir('/tmp', function(error) {
+    fs.mkdir('/tmp', function (error) {
       expect(error).not.to.exist;
-      if(error) throw error;
+      if (error) throw error;
 
-      fs.stat('/tmp', function(error, stats) {
+      fs.stat('/tmp', function (error, stats) {
         expect(error).not.to.exist;
         expect(stats).to.exist;
         expect(stats.type).to.equal('DIRECTORY');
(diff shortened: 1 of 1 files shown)


jespirit:
In general, it's a better idea to separate the whitespace-only changes into a different commit because the mkdir promises tests are unrelated to the whitespace changes. However, in your particular case it's just a matter of programming style, some like to enter a space and some don't between `function () {` for anonymous functions. So you could just leave as is.

klymenkoo:
@jespirit I definitely agree with you. I didn't realize the whitespaces got in as I commited file right after I used VSCode "Format document" feature.

tests/spec/fs.mkdir.spec.js
```diff
@@ -46,3 +46,51 @@ describe('fs.mkdir', function() {
...
+
+  it('should be a function', function () {
+    var fs = util.fs();
+    expect(fs.promises.mkdir).to.be.a('function');
+  });
+
+  it('should return an error if part of the parent path does not exist', function (done) {
+    var fs = util.fs();
+
+    fs.promises.mkdir('/tmp/mydir', function (error) {
```
humphd: This is mixing callbacks and promises.  You'd need to change to something like this:

```js
fs.promises.mkdir('/tmp/mydir', () =>
  return fs.promises
    .mkdir('/tmp/mydir')
    .catch(error => {
      expect(error).to.exist);
    });
);
```

tests/spec/fs.mkdir.spec.js
```diff
@@ -46,3 +46,51 @@ describe('fs.mkdir', function() {
...
+      expect(error).to.exist;
+      expect(error.code).to.equal('ENOENT');
+    }).finally(done);
+  });
+
+  it('should return an error if the path already exists', function (done) {
+    var fs = util.fs();
+
+    fs.promises.mkdir('/', function (error) {
```
humphd: Same issue here.  You want to return your Promise, and use then/catch vs. a callback.

tests/spec/fs.mkdir.spec.js
```diff
@@ -46,3 +46,51 @@ describe('fs.mkdir', function() {
...
+      expect(error).to.exist;
+      expect(error.code).to.equal('EEXIST');
+    }).finally(done);
+  });
+
+  it('should make a new directory', function (done) {
+    var fs = util.fs();
+
+    fs.promises.mkdir('/tmp', function (error) {
```
humphd: And here, same again.  Switch to promises.

klymenkoo:
@humphd thank you for the feedback! Take a look at the new commit with updated code.

klymenkoo:
@humphd done ✅

tests/spec/fs.mkdir.spec.js
```diff
@@ -46,3 +46,61 @@ describe('fs.mkdir', function() {
...
+          expect(error).to.exist;
+          expect(error.code).to.equal('EEXIST');
+        });
+    });
+  });
+
+  it('should make a new directory', function () {
+    var fs = util.fs();
+
+    fs.promises.mkdir('/tmp', () => {
```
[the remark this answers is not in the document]
klymenkoo: @humphd actually, `fs.stat` does not return `Promise`, so I don't think that I can move its callback to `then`

klymenkoo:
@humphd, I feel like the latest code update should be final. Take a look, and let me know if there is anything left to change.

humphd:
@klymenkoo can you rebase this please?

humphd:
@klymenkoo can you check again for me, https://github.com/filerjs/filer/pull/415/conflicts says it is still conflicting.

klymenkoo:
@humphd, this is the message I got from git after rebase:
 `Successfully rebased and updated refs/heads/issue-407.`
Do you know how can I reproduce this conflict on my local machine?

humphd:
Did you push your branch again with `-f`?

klymenkoo:
@humphd yes.

humphd:
Just to confirm, here is what needs to happen (maybe you did this, in case not):

```
git checkout master
git pull upstream master
git checkout issue-407
git rebase master
...fix this merge conflict in tests/spec/fs.mkdir.spec.js...
git add tests/spec/fs.mkdir.spec.js
git rebase --continue
git push origin issue-407 -f
```

klymenkoo:
@humphd I caught my mistake. It should be good now.

PR #416, opened 2018-09-20
fix issue#392-- add test for promises.rename() to rename existing file

### LINKED ISSUE #392: Add a test for promises.rename() for renaming an existing file
Current tests in [fs.rename.spec.js] (https://github.com/filerjs/filer/blob/develop/tests/spec/fs.rename.spec.js) do not include a test to check if fs.promises.rename() successfully renames an existing file.

The new test could be based on logic for the equivalent existing Async test: 
```js
 it('should rename an existing file', function(done) { //...
```
and possibly logic found for rename in an existing  [Node.js test] (https://github.com/nodejs/node/blob/fdf829eea48206d0a078dba3a0bf9f77c6acb619/test/parallel/test-fs-promises.js)

I'd like to make this change if it seems reasonable, and add it to the existing test suite in fs.rename.spec.js

Michael-Overall:
Adds test for promises.rename() to tests/spec/fs.rename,spec.js to check renaming an existing file.

Closely mimics logic for an existing test that checks if the the Async rename() function renames an existing file (so no fs.promises functions are used for open(), close(), or stat() ).

Closes #392 

/cc @humphd

### DIFF
--- a/tests/spec/fs.rename.spec.js
+++ b/tests/spec/fs.rename.spec.js
@@ -151,4 +151,44 @@ describe('fs.rename', function() {
       });
     });
   });
+
+  it('(promise version) should rename an existing file', function(done) {
+    var complete1 = false;
+    var complete2 = false;
+    var fs = util.fs();
+
+    function maybeDone() {
+      if(complete1 && complete2) {
+        done();
+      }
+    }
+
+    fs.open('/myfile', 'w+', function(error, fd) {
+      if(error) throw error;
+
+      fs.close(fd, function(error) {
+        if(error) throw error;
+
+        fs.promises.rename('/myfile', '/myotherfile').then(
+          function(){
+            fs.stat('/myfile', function(error, result) {
+              expect(error).to.exist;
+              expect(result).not.to.exist;
+              complete1 = true;
+              maybeDone();
+            });
+      
+            fs.stat('/myotherfile', function(error, result) {
+              expect(error).not.to.exist;
+              expect(result.nlinks).to.equal(1);
+              complete2 = true;
+              maybeDone();
+            });
+          },
+          function(error){throw error;}
+        );
+        
+      });
+    });
+  });
 });


tests/spec/fs.rename.spec.js
```diff
@@ -151,4 +151,44 @@ describe('fs.rename', function() {
...
+      }
+    }
+
+    fs.open('/myfile', 'w+', function(error, fd) {
+      if(error) throw error;
+
+      fs.close(fd, function(error) {
+        if(error) throw error;
+
+        fs.promises.rename('/myfile', '/myotherfile').then(
```
humphd: I think it's confusing that you're mixing callbacks with promises in this test.  Since you're testing the promise version, can you switch all these over to use promises instead?

tests/spec/fs.rename.spec.js
```diff
@@ -151,4 +151,44 @@ describe('fs.rename', function() {
...
+          function(){
```
humphd: With promises, we should be able to get rid of this hack, and instead use [`Promise.all()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all)

tests/spec/fs.rename.spec.js
Michael-Overall: Replaced all async functions with their promise equivalents (commit:  5abefb6c6a467eb0271caddc0685760427fd57bc). I also noticed later that my IDE added some spaces after all occurrences of "function" and "if" so this was corrected in 98611f4d7669488d365e8f07f1b43f582a52da6f

tests/spec/fs.rename.spec.js
Michael-Overall: Implemented Promise.all() version in 5abefb6c6a467eb0271caddc0685760427fd57bc. Using promises certainly cleaned up the code.

tests/spec/fs.rename.spec.js
```diff
@@ -151,4 +151,35 @@ describe('fs.rename', function() {
       });
     });
   });
+
+  it('(promise version) should rename an existing file', function(done) {
+    var complete1 = false;
+    var complete2 = false;
+    var fs = util.fs();
+
+    function maybeDone() {
```
humphd: Instead of doing things this way, with a `done` callback and a bunch of booleans to track what is and isn't finished, we can just rely on Promises to do it.  Change your test code like this:

1. Remove the `done` callback arg on line 155.  We won't use it (see 2.)
1. Add a `return` to the front of line 167 below, so you are returning your promise.  This will tell it when the test is done.
1. See my inline changes below

tests/spec/fs.rename.spec.js
```diff
@@ -151,4 +151,35 @@ describe('fs.rename', function() {
...
+      if(complete1 && complete2) {
+        done();
+      }
+    }
+    
+  
+    fs.promises.open('/myfile', 'w+')
+      .then((fd)=>fs.promises.close(fd))
+      .then(fs.promises.rename('/myfile', '/myotherfile'))
```
humphd: Should be:

```js
.then(() => fs.promises.rename('/myfile', '/myotherfile'))
```

tests/spec/fs.rename.spec.js
```diff
@@ -151,4 +151,35 @@ describe('fs.rename', function() {
...
+      .then(Promise.all(fs.promises.stat('/myfile')
```
humphd:
```js
.then(() => fs.promises.stat('/myfile')
```

tests/spec/fs.rename.spec.js
```diff
@@ -151,4 +151,35 @@ describe('fs.rename', function() {
...
+        .then( (result)=> expect(result).not.to.exist, (error) => expect(error).to.exist)
```
humphd:
```js
.catch(error => expect(error).to.exist)
.then(() => fs.promises.stat('/myotherfile'))
...and so on
```

PR #432, opened 2018-09-21
Fix #423: added proimse support for fs.stat

### LINKED ISSUE #423: Add test for fs.promises.stat() when path exists
The fs.spec.stat file needs a test using promise support when the path exists. The version using callbacks can be seen here.
```js
  it('should return a stat object if path exists', function(done) {
    var fs = util.fs();

    fs.stat('/', function(error, result) {
      expect(error).not.to.exist;
      expect(result).to.exist;

      expect(result['node']).to.be.a('string');
      expect(result['dev']).to.equal(fs.name);
      expect(result['size']).to.be.a('number');
      expect(result['nlinks']).to.be.a('number');
      expect(result['atime']).to.be.a('number');
      expect(result['mtime']).to.be.a('number');
      expect(result['ctime']).to.be.a('number');
      expect(result['type']).to.equal('DIRECTORY');

      done();
    });
});
```
Is it okay for me to work on add a promise version of the above code?

rdittrich97:
Made a test for fs.stat.promise when the path exists.

### DIFF
--- a/tests/spec/fs.stat.spec.js
+++ b/tests/spec/fs.stat.spec.js
@@ -91,21 +91,21 @@ describe('fs.stat', function() {
       });
     });
   });
-});
 
-describe('fs.promises.stat', function() {
-  beforeEach(util.setup);
-  afterEach(util.cleanup);
-
-  it('should be a function', function() {
+  it('(promise) should be a function', function() {
     var fs = util.fs();
     expect(fs.promises.stat).to.be.a('function');
   });
   
-  it('should return a stat object if file exists', function(done){
+  it('should return a promise', function() {
+    var fs = util.fs();
+    expect(fs.promises.stat()).to.be.a('Promise');
+  });
+  
+  it('(promise) should return a stat object if file exists', function(){
     var fs = util.fs();
 
-    fs.promises.stat('/')
+    return fs.promises.stat('/')
       .then(result => { 
         expect(result).to.exist;
         expect(result['node']).to.be.a('string');
@@ -116,7 +116,6 @@ describe('fs.promises.stat', function() {
         expect(result['mtime']).to.be.a('number');
         expect(result['ctime']).to.be.a('number');
         expect(result['type']).to.equal('DIRECTORY');
-        done();
       })
       .catch(error => expect(error).not.to.exist);
   });
--- a/tests/spec/fs.stat.spec.js
+++ b/tests/spec/fs.stat.spec.js
@@ -102,21 +102,21 @@ describe('fs.stat', function() {
     expect(fs.promises.stat()).to.be.a('Promise');
   });
   
-  it('(promise) should return a stat object if file exists', function(){
+  it('(promise) should return a stat object if file exists', function() {
     var fs = util.fs();
 
-    return fs.promises.stat('/')
+    return fs.promises
+      .stat('/')
       .then(result => { 
         expect(result).to.exist;
-        expect(result['node']).to.be.a('string');
-        expect(result['dev']).to.equal(fs.name);
-        expect(result['size']).to.be.a('number');
-        expect(result['nlinks']).to.be.a('number');
-        expect(result['atime']).to.be.a('number');
-        expect(result['mtime']).to.be.a('number');
-        expect(result['ctime']).to.be.a('number');
-        expect(result['type']).to.equal('DIRECTORY');
-      })
-      .catch(error => expect(error).not.to.exist);
+        expect(result.node).to.be.a('string');
+        expect(result.dev).to.equal(fs.name);
+        expect(result.size).to.be.a('number');
+        expect(result.nlinks).to.be.a('number');
+        expect(result.atime).to.be.a('number');
+        expect(result.mtime).to.be.a('number');
+        expect(result.ctime).to.be.a('number');
+        expect(result.type).to.equal('DIRECTORY');
+      }); 
   });
 });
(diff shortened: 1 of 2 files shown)


tests/spec/fs.stat.spec.js
```diff
@@ -92,3 +92,32 @@ describe('fs.stat', function() {
...
+describe('fs.promises.stat', function() {
+  beforeEach(util.setup);
+  afterEach(util.cleanup);
+
+  it('should be a function', function() {
+    var fs = util.fs();
+    expect(fs.promises.stat).to.be.a('function');
+  });
+  
+  it('should return a stat object if file exists', function(done){
```
y2s82: To make this test more promise-friendly, please remove the `done` from the callback function arguments.

tests/spec/fs.stat.spec.js
```diff
@@ -92,3 +92,32 @@ describe('fs.stat', function() {
...
+    var fs = util.fs();
+
+    fs.promises.stat('/')
```
y2s82: To make this test more promise-friendly, please return the return of this function.

tests/spec/fs.stat.spec.js
```diff
@@ -92,3 +92,32 @@ describe('fs.stat', function() {
...
+        expect(result).to.exist;
+        expect(result['node']).to.be.a('string');
+        expect(result['dev']).to.equal(fs.name);
+        expect(result['size']).to.be.a('number');
+        expect(result['nlinks']).to.be.a('number');
+        expect(result['atime']).to.be.a('number');
+        expect(result['mtime']).to.be.a('number');
+        expect(result['ctime']).to.be.a('number');
+        expect(result['type']).to.equal('DIRECTORY');
+        done();
```
y2s82: To make this test more promise-friendly, please remove the unnecessary `done();`.

tests/spec/fs.stat.spec.js
```diff
@@ -92,3 +92,32 @@ describe('fs.stat', function() {
     });
   });
 });
+
+describe('fs.promises.stat', function() {
```
y2s82: Move the following test into the `describe('fs.stat'...)` block.  This makes it easier to add more related tests.

rdittrich97:
Thank you for your feedback @y2s82

tests/spec/fs.stat.spec.js
```diff
@@ -91,4 +91,32 @@ describe('fs.stat', function() {
...
+    expect(fs.promises.stat()).to.be.a('Promise');
+  });
+  
+  it('(promise) should return a stat object if file exists', function(){
+    var fs = util.fs();
+
+    return fs.promises.stat('/')
+      .then(result => { 
+        expect(result).to.exist;
+        expect(result['node']).to.be.a('string');
```
humphd: Doing `result['node']` vs. `result.node` is fine, but the latter is more common.  If we're doing `[]` indexing into the object above, I can see why you copied the same style.  In general, I would prefer `result.node`, unless the key name is reserved, or being assembled as a variable.  Same below.

tests/spec/fs.stat.spec.js
```diff
@@ -91,4 +91,32 @@ describe('fs.stat', function() {
...
+    var fs = util.fs();
+    expect(fs.promises.stat).to.be.a('function');
+  });
+  
+  it('should return a promise', function() {
+    var fs = util.fs();
+    expect(fs.promises.stat()).to.be.a('Promise');
+  });
+  
+  it('(promise) should return a stat object if file exists', function(){
```
humphd: style nit: `, function() { ` add a space between your `){`

tests/spec/fs.stat.spec.js
```diff
@@ -91,4 +91,32 @@ describe('fs.stat', function() {
...
+        expect(result['node']).to.be.a('string');
+        expect(result['dev']).to.equal(fs.name);
+        expect(result['size']).to.be.a('number');
+        expect(result['nlinks']).to.be.a('number');
+        expect(result['atime']).to.be.a('number');
+        expect(result['mtime']).to.be.a('number');
+        expect(result['ctime']).to.be.a('number');
+        expect(result['type']).to.equal('DIRECTORY');
+      })
+      .catch(error => expect(error).not.to.exist);
```
humphd: I don't think this is necessary.  Unless you are expecting an `error` case, which you aren't here, you can just leave the `catch` off and let the test handle it: since you return your `Promise` on line 108, the encapsulating test (`it()` function) will receive the error and report a failure.

tests/spec/fs.stat.spec.js
```diff
@@ -91,4 +91,32 @@ describe('fs.stat', function() {
...
+  
+  it('should return a promise', function() {
+    var fs = util.fs();
+    expect(fs.promises.stat()).to.be.a('Promise');
+  });
+  
+  it('(promise) should return a stat object if file exists', function(){
+    var fs = util.fs();
+
+    return fs.promises.stat('/')
```
humphd: Style nit: drop your function calls down to all have the same indent (e.g. `stat`, `then`), to set them out visually.

```js
return fs.promises
  .stat('/')
  .then(result => { ...
```

rdittrich97:
Thank you for the comments, the changes you requested were implemented.

tests/spec/fs.stat.spec.js
```diff
@@ -91,4 +91,33 @@ describe('fs.stat', function() {
       });
     });
   });
+
+  it('(promise) should be a function', function() {
+    var fs = util.fs();
```
struong9: Maybe consider doing this to make it more clear what is being tested.

rdittrich97: Could you please elaborate on the necessity of this change?

struong9: As the rest of the tests in this file are for fs, it's just better to make it more obvious you're testing for fsPromises.

PR #437, opened 2018-09-21
Fix #435: add a test for fsPromise.unlink() to delete a file that doe…

### LINKED ISSUE #435: Add a test for fsPromise.unlink() to delete a file that does not exist.
I want to add a test for `fsPromise.unlink()` to delete a file that does not exist. I like to work on this issue. Would you please assign me to this test?

susantruong:
I wrote a test for `fsPromise.unlink()` to delete a file that does not exist. If the file does not exist, it can not be opened, and throw an error. https://github.com/filerjs/filer/issues/435

### DIFF
--- a/tests/spec/fs.unlink.spec.js
+++ b/tests/spec/fs.unlink.spec.js
@@ -105,7 +105,6 @@ describe('fs.unlink', function() {
   });
 });
 
-
 describe('fs.promises.unlink', function () {
   beforeEach(util.setup);
   afterEach(util.cleanup);
@@ -113,5 +112,15 @@ describe('fs.promises.unlink', function () {
   it('should be a function', function () {
     var fs = util.fs();
     expect(fs.promises.unlink).to.be.a('function');
+  });
+
+  it('should return an error if trying to delete a file that does not exist', function() {
+    var fsPromises = util.fs().promises;
+
+    return fsPromises.unlink('/myFile')
+      .catch(error => {
+        expect(error).to.exist;
+        expect(error.code).to.equal('ENOENT');
+      });
   });
 });


tests/spec/fs.unlink.spec.js
```diff
@@ -104,3 +104,22 @@ describe('fs.unlink', function() {
...
   });
 });
+
+/**
+ * fsPromises tests
+ */
+describe('fsPromises.stat', function() {
+  beforeEach(util.setup);
+  afterEach(util.cleanup);
+  it('should return an error if trying to delete a file that does not exist', function(done) {
```
y2s82: To make code more promise-friendly, please remove `done` from the argument

tests/spec/fs.unlink.spec.js
```diff
@@ -104,3 +104,22 @@ describe('fs.unlink', function() {
...
+    var fsPromises = util.fs().promises;
+
+    fsPromises.unlink('/myFile')
```
y2s82: To make this more promise-friendly, please return this function.

tests/spec/fs.unlink.spec.js
```diff
@@ -104,3 +104,22 @@ describe('fs.unlink', function() {
...
+      .then(result => expect(result).not.to.exist)
+      .catch(error => {
+        expect(error).to.exist;
+        expect(error.code).to.equal('ENOENT');
+      })
+      .then(() => done());
```
y2s82: To make this code more promise-friendly, please remove this line as `done()` is no longer needed.

tests/spec/fs.unlink.spec.js
```diff
@@ -104,3 +104,22 @@ describe('fs.unlink', function() {
...
+ * fsPromises tests
+ */
+describe('fsPromises.stat', function() {
+  beforeEach(util.setup);
+  afterEach(util.cleanup);
+  it('should return an error if trying to delete a file that does not exist', function(done) {
+    var fsPromises = util.fs().promises;
+
+    fsPromises.unlink('/myFile')
+      .then(result => expect(result).not.to.exist)
```
y2s82: This line is also unneeded in a promise-based test, so please remove this line.

susantruong: @humphd Would you please help me on this issue? Do I need to include it?

tests/spec/fs.unlink.spec.js
susantruong: thank you for your suggestion :)

tests/spec/fs.unlink.spec.js
susantruong: thank you for your suggestion :)

tests/spec/fs.unlink.spec.js
susantruong: thank you for your suggestion :)

tests/spec/fs.unlink.spec.js
humphd: I think you can remove it, yes.  It's not wrong to have it, but it's not doing anything we care about.  With callbacks, you need to do this, because the callback gets called in both the error and success cases, so you have to examine state for both.  With Promises, only one of the two conditions will happen.

humphd:
@susantruong This looks good, but there's now a merge conflict (not your fault) because I've merged some other students' code.  Let's fix this together in class later this week.

susantruong:
@humphd I have made a change that you and my classmate recommended. Would you please review it? Thanks.

PR #439, opened 2018-09-22
Fix #431: Add test for fsPromises.unlink() when trying to remove a directory.

### LINKED ISSUE #431: Add test for fsPromises.unlink() when trying to remove a directory.
Currently the [unlink() tests](https://github.com/filerjs/filer/blob/master/tests/spec/fs.unlink.spec.js) do not contain a test for [fsPromises.unlink()](https://nodejs.org/docs/latest/api/fs.html#fs_fspromises_unlink_path). The code for a test like this would be very similar to the ones already existing for fs.unlink()

```js
it('should not unlink directories', function (done) {
  var fs = util.fs();

  fs.mkdir('/mydir', function (error) {
    if(error) throw error;

    fs.unlink('/mydir', function (error) {
      expect(error).to.exist;
      expect(error.code).to.equal('EPERM');

      fs.stat('/mydir', function (error, stats) {
        expect(error).not.to.exist;
        expect(stats).to.exist;
        expect(stats.type).to.equal('DIRECTORY');
        done();
      });
    });
  });
});
```

I would like to be assigned this issue if possible.

jagmeetb:
Hi, this is my attempt at adding an fsPromises.unlink() test. I would appreciate any feedback!


dnguneratne:
Good effort but there's room for improvement.

1. Chain the different `fs.*` method using `.then` to ensure that they execute in order.
2. In this test there are errors that you expect to happen as part of the test (should not unlink directories), and errors that may happen. 

	a. For errors that may happen you can use a single `.catch` at the end of the promise chain.
e.g:

	```js
	fs.doSomething
		.then(() => {..})
		.then(() => {..})
		.then(() => {..})
		.catch(err => { throw err; });
	```

	If an error occurs inside any of the `.then` statements it will be caught in the closest `.catch` in the same level. In this case, the last `.catch` statement.  

	b. For errors that you expect, use a `.catch` instead of a `.then`. The first argument of `.then` represents the value returned from the promise if it was *successfully* executed. Consider this section;

	```js
	fs.unlink('/mydir')
      .then(
        error => { // error is the value unlink returns if it executed successfully. Which in this case will not happen.
		  // Therefore, the following assertions will never run.
		  // console.log('here'); // Will not appear on console
          expect(error).to.exist;
          expect(error.code).to.equal('EPERM');
        });
	```
	If we refactor this block to;
	```js
	fs.unlink('/mydir')
      .catch(error => {
		// console.log('here'); // Will appear on console
        expect(error).to.exist;
        expect(error.code).to.equal('EPERM');
      });
	```	
	The assertions will run. You can also test this by purposefully failing an assertion.
	e.g: Change `expect(error).to.exist;` to `expect(error).not.to.exist;`, and the test should fail.

3. Starting v1.18.0 Mocha has added support for promises, meaning we can get rid of the `done` callback. Return promises to the mocha test and it will recognize the test as async. While you are at it, to improve conciseness, you can also remove the `function` keyword in favor of the arrow functions.

	```js
	it('should not unlink directories', () => { // No function or done keyword.
		var fs = util.fs().promises;

		return fs.mkdir('/mydir') // Notice the return
			.then(() => { 
				return fs.unlink('/mydir') // Return promise functions
					.catch(error => {});
			}).then(...
		
		// .then(() => done()); Not required.
	```
	See PR #412 for an example.

4. Formatting improvements - move the `.then` callback arguments inline with `.then` method:
	
	```js
	fs.stat('/mydir')
      .then(stats => {
          expect(stats).to.exist;
          expect(stats.type).to.equal('DIRECTORY');
       })
	```

tests/spec/fs.unlink.spec.js
```diff
@@ -104,3 +104,41 @@ describe('fs.unlink', function() {
...
+
+describe('fsPromises.unlink', function() {
+  beforeEach(util.setup);
+  afterEach(util.cleanup);
+
+  it('should not unlink directories', function (done) {
+    var fs = util.fs().promises;
+
+    fs.mkdir('/mydir')
+      .then(
```
dnguneratne: See point 2a on using `.catch`

tests/spec/fs.unlink.spec.js
```diff
@@ -104,3 +104,41 @@ describe('fs.unlink', function() {
...
+        error => {
+          if(error) throw error;
+        });
+
+    fs.unlink('/mydir')
+      .then(
```
dnguneratne: See point 2b about `.then` arguments

tests/spec/fs.unlink.spec.js
```diff
@@ -104,3 +104,41 @@ describe('fs.unlink', function() {
...
+        error => {
+          expect(error).to.exist;
+          expect(error.code).to.equal('EPERM');
+        });
+
+    fs.stat('/mydir')
+      .then(
```
dnguneratne: See point 4 on formatting improvements.

tests/spec/fs.unlink.spec.js
```diff
@@ -104,3 +104,41 @@ describe('fs.unlink', function() {
...
+
+/**
+ * fsPromises tests
+ */
+
+describe('fsPromises.unlink', function() {
+  beforeEach(util.setup);
+  afterEach(util.cleanup);
+
+  it('should not unlink directories', function (done) {
```
dnguneratne: See point 3 - leverage Mocha's built-in promise support

jagmeetb:
@dnguneratne thanks for the reply, I'll implement your suggestions.

tests/spec/fs.unlink.spec.js
```diff
@@ -104,3 +104,41 @@ describe('fs.unlink', function() {
...
+    var fs = util.fs().promises;
+
+    fs.mkdir('/mydir')
+      .then(
+        error => {
+          if(error) throw error;
+        });
+
+    fs.unlink('/mydir')
```
dnguneratne: See point 1 - Chain function calls with `.then` to ensure proper execution order.

jagmeetb:
@dnguneratne I've implemented your suggestions, and think everything should be good now.

tests/spec/fs.unlink.spec.js
```diff
@@ -104,3 +104,37 @@ describe('fs.unlink', function() {
...
+            expect(error.code).to.equal('EPERM');
+          });
+      })
+      .then(() => {
+        return fs.stat('/mydir')
+          .then(stats => {
+            expect(stats).to.exist;
+            expect(stats.type).to.equal('DIRECTORY');
+          })
+          .then( error => {
```
dnguneratne: 1. This `.then` block is not needed. This is an error that may happen, but not part of the actual test (for example, in the case of `fs.unlink('/mydir')` (line 121), for the test to pass an error must exist. Which is why you used a `.catch`). If an error occurred during `fs.stat('/mydir')` (line 128), it'll be caught in the `.catch` on line 137. Once that happens the test will fail anyway, so you don't need to explicitly test for an error to not exist.

2. Consider the following block:
```js
	return fs.stat('/mydir')
   		.then(stats => {...})
		.then( error => {..})
```
The first argument of `.then` is the return value if the promise was **successfully** resolved. `error` in `.then( error => {..})` would represent the return value of the `.then` block before it if it was a promise, and that promise was successfully resolved. However, the `.then` block (line 128) contains chai assertions. Therefore, `.then( error => {..})` part is not needed.

tests/spec/fs.unlink.spec.js
```diff
@@ -104,3 +104,37 @@ describe('fs.unlink', function() {
...
+            expect(error).not.to.exist;
+          });
+      })
+      .catch(error => {
```
dnguneratne: You can simplify this to a one liner with:
.catch(error => { throw error; });

jagmeetb:
I've removed the `.then ({ ... })` block you mentioned and minified the `catch` block to 1 line.

tests/spec/fs.unlink.spec.js
```diff
@@ -104,3 +104,33 @@ describe('fs.unlink', function() {
...
+/**
+ * fsPromises tests
+ */
+
+describe('fsPromises.unlink', function() {
+  beforeEach(util.setup);
+  afterEach(util.cleanup);
+
+  it('should not unlink directories', () => {
+    var fs = util.fs().promises;
```
struong9: Maybe consider naming the variable fsPromises to make it more clear what is being tested.

struong9: Maybe consider naming the variable fsPromises to make it more clear what is being tested.

tests/spec/fs.unlink.spec.js
```diff
@@ -104,3 +104,33 @@ describe('fs.unlink', function() {
...
+
+    return fs.mkdir('/mydir')
+      .then(() => {
+        return fs.unlink('/mydir')
```
humphd: A tip: when you use fat arrow `=>` functions like this, if the only thing your function does is to return a value (e.g., it's an expression), you can get rid of the `{ return...}` and do this instead:

```js
return fs.mkdir('/mydir')
  .then(() => fs.unlink('/mydir'))
```

What you have now is fine, just an FYI for later.  Nice work.

humphd:
@jagmeetb great work, and @dnguneratne, an epic review!

This is good to go as soon as we fix the merge conflict.  Let's discuss in class tomorrow.

humphd:
I've rebased and landed this in https://github.com/filerjs/filer/commit/1abcb0369b925324599bd7eab8a9d71b31076f7e.

PR #440, opened 2018-09-22
Issue 428 problem with using stat and lstat

Janice0218:
I changed the code, using 'lstat' instead of 'stat',
[changed code](https://github.com/Janice0218/filer/commit/e73099ce1a32db75e19c6157431a7263ab80b8c6)
`it('should return an error if path does not exist', function(done) {

     var fs = util.fs();

     //'stat' should be 'lstat'

     fs.stat('/tmp', function(error, result) {

     expect(error).to.exist;

      expect(result).not.to.exist;

     done();
     });
});`

because i think stat follows symlinks. When given a path that is a symlink, it returns the stat of the target of the symlink.
lstat doesn't follow symlinks.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,8 +10,8 @@
       "integrity": "sha512-bPHp6Ji8b41szTOcaP63VlnbbO5Ny6dwAATtY6JTjh5N2OLrb5Qk/Th5cRkRQhkWCt+EJsYrNB0MiL+Gpn6e3g==",
       "dev": true,
       "requires": {
-        "call-me-maybe": "^1.0.1",
-        "glob-to-regexp": "^0.3.0"
+        "call-me-maybe": "1.0.1",
+        "glob-to-regexp": "0.3.0"
       }
     },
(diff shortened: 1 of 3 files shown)


humphd:
@Janice0218 I think that we do want to use `lstat` and `stat` in these tests, since they do different things.  With `stat` we are getting stats for the file pointed to by `path`.  With `lstat` we are also allowing a symbolic link, and then we are stating the  link itself, not the file that it refers to.  It's less common to use `lstat`, but it's something we sometimes need to do (e.g., when we care about the link file itself vs. the thing it points to).  A symbolic link is really just a file with a filename as its contents--the file to which it links.

Based on the above, I think this PR isn't needed.  Can you confirm that I'm not misunderstanding?  Either way, this PR looks good from a process point of view, and you can use it for your lab.  I would just like to see you work on a PR that we can merge.

humphd:
I'm going to close this, as the calls to `lstat` are right in this case.  Thanks for opening a PR for it, it was good to check that we were doing the right thing.

Issue #441, opened 2018-09-22
Add mkdtemp method


AlexeiGitH:
Node.js File System module has a method [mkdtemp](https://nodejs.org/api/fs.html#fs_fs_mkdtemp_prefix_options_callback), used to create a temporary directory. 

I would like to implement this method in Filer, before adding tests methods, mentioned in issue #394 

`
  fs.mkdtemp(prefix, callback);
`

Here prefix is a name for a temporary directory. 6 random characters will be appended to the name to make sure directory is unique.
Error and path to created directory will be passed to a callback.

All temp directories will be created in /tmp.

humphd:
Sounds good.  One thing to note: I don't think `node` assumes you'll always put your temp files in `/tmp`.  My reading of the docs is that you specify a prefix, and it uses that.  So for example you could do: `mkdtemp('/tmp/', callback)` and it will create `/tmp/ay3ri1`.  But you could also use `mkdtemp('/foo/tmp-', callback)` and get `/foo/tmp-ay3ri1`.  We should confirm this so we match what node does.

AlexeiGitH:
I assumed that all new tmp directories will be created in /tmp.

I will test how node handles cases you mentioned, and will update mkdtemp method accordingly.

humphd:
Finished in https://github.com/filerjs/filer/commit/0eb7066a861d9d4506a4b2b2bb10865379aab82c

PR #443, opened 2018-09-22
Resolves issue #441. Added mkdtemp method. 

AlexeiGitH:
New method **mkdtemp** has been added to 'src/filesystem/implementation.js'.
Interface for created method added to 'src/filesystem/interface.js'.

### DIFF
--- a/README.md
+++ b/README.md
@@ -306,7 +306,7 @@ var fs = new Filer.FileSystem();
 * [fs.mknod(path, mode, callback)](#mknod)
 * [fs.rmdir(path, callback)](#rmdir)
 * [fs.mkdir(path, [mode], callback)](#mkdir)
-* [fs.mkdtemp(path, [mode], callback)](#mkdtemp)
+* [fs.mkdtemp(path, [options], callback)](#mkdtemp)
 * [fs.readdir(path, callback)](#readdir)
 * [fs.close(fd, callback)](#close)
 * [fs.open(path, flags, [mode], callback)](#open)
@@ -676,10 +676,9 @@ fs.mkdir('/home', function(err) {
 
 #### fs.mkdtemp(prefix, options, callback)<a name="mkdtemp"></a>
 
+Makes a temporary directory with prefix supplied in `path` argument. Method will append six random characters directly to the prefix.
 
-Makes a temporary directory with prefix supplied in `path` argument. Method will append the six randomly selected characters directly to the prefix.
-
-Asynchronous. Callback gets `(error, path)`, where path is a path to created directory.
+Asynchronous. Callback gets `(error, path)`, where path is the path to the created directory.
 
 NOTE: Filer allows for, but ignores the optional `options` argument used in node.js.
 
@@ -687,7 +686,7 @@ Example:
 
 ```javascript
 // Create tmp directory with prefix foo
-fs.mkdtemp("foo-", function (error, path) {
+fs.mkdtemp("/foo-", function (error, path) {
     //A new folder foo-xxxxxx will be created. Path contains a path to created folder.    
 });
 
@@ -695,9 +694,7 @@ fs.mkdtemp("/myDir/tmp", function (error, path) {
     //Will create a new folder tmpxxxxxx inside myDir directory. 
     //Will throw error if myDir does not exist    
 });
-
 ```
-
 #### fs.readdir(path, callback)<a name="readdir"></a>
 
 Reads the contents of a directory. Asynchronous [readdir(3)](http://pubs.opengroup.org/onlinepubs/009695399/functions/readdir.html).
--- a/README.md
+++ b/README.md
@@ -306,6 +306,7 @@ var fs = new Filer.FileSystem();
 * [fs.mknod(path, mode, callback)](#mknod)
 * [fs.rmdir(path, callback)](#rmdir)
 * [fs.mkdir(path, [mode], callback)](#mkdir)
+* [fs.mkdtemp(path, [mode], callback)](#mkdtemp)
 * [fs.readdir(path, callback)](#readdir)
 * [fs.close(fd, callback)](#close)
 * [fs.open(path, flags, [mode], callback)](#open)
@@ -671,6 +672,30 @@ fs.mkdir('/home', function(err) {
     // directory /home/carl now exists
   });
 });
+```
+
+#### fs.mkdtemp(prefix, options, callback)<a name="mkdtemp"></a>
+
+
+Makes a temporary directory with prefix supplied in `path` argument. Method will append the six randomly selected characters directly to the prefix.
+
+Asynchronous. Callback gets `(error, path)`, where path is a path to created directory.
+
+NOTE: Filer allows for, but ignores the optional `options` argument used in node.js.
+
+Example:
+
+```javascript
+// Create tmp directory with prefix foo
+fs.mkdtemp("foo-", function (error, path) {
+    //A new folder foo-xxxxxx will be created. Path contains a path to created folder.    
+});
+
+fs.mkdtemp("/myDir/tmp", function (error, path) {
+    //Will create a new folder tmpxxxxxx inside myDir directory. 
+    //Will throw error if myDir does not exist    
+});
+
 ```
 
 #### fs.readdir(path, callback)<a name="readdir"></a>
--- a/src/filesystem/implementation.js
+++ b/src/filesystem/implementation.js
@@ -1681,7 +1681,7 @@ function mkdir(fs, context, path, mode, callback) {
 function mkdtemp(fs, context, prefix, options, callback) { 
   callback = arguments[arguments.length - 1];
 
-  // this function is ised to generate a random character set to append 
+  // this function is used to generate a random character set to append 
   // to tmp dir name to make sure it is unique
   function generateRandom() {
     return 'xxxxxx'.replace(/[xy]/g, function(c) {
--- a/src/filesystem/implementation.js
+++ b/src/filesystem/implementation.js
@@ -1687,19 +1687,19 @@ function mkdtemp(fs, context, prefix, options, callback) {
     return 'xxxxxx'.replace(/[xy]/g, function(c) {
       var r = Math.random()*16|0, v = c == 'x' ? r : (r&0x3|0x8);
       return v.toString(16);
-    })
+    });
   }
   if (prefix) {
-    var path = "/"+prefix+generateRandom();
+    var path = '/'+prefix+generateRandom();
     make_directory(context, path, function(error) {
       if (error) {
-        callback(error, path)
+        callback(error, path);
       } else {
         callback(null, path);
       }
     });
   } else {
-    callback(new Error('filename prefix is required'), prefix)
+    callback(new Error('filename prefix is required'), prefix);
   }  
 }
 
--- a/src/filesystem/implementation.js
+++ b/src/filesystem/implementation.js
@@ -1678,6 +1678,55 @@ function mkdir(fs, context, path, mode, callback) {
   make_directory(context, path, callback);
 }
 
+function mkdtemp(fs, context, prefix, callback) { 
+  callback = arguments[arguments.length - 1];
+
+  // this function is ised to generate a random character set to append 
+  // to tmp dir name to make sure it is unique
+  function generateRandom() {
+    return 'xxxxxx'.replace(/[xy]/g, function(c) {
+      var r = Math.random()*16|0, v = c == 'x' ? r : (r&0x3|0x8);
+      return v.toString(16);
+    })
+  }
+
+  var tmpDir = "/tmp";      //tmp dir in root
+  var tmpDirExists = false;   //flag to check if already created
+  
+  if (!prefix) {
+    prefix = generateRandom(); //if user did not specify prefix
+  }
+  prefix = prefix + "-" + generateRandom();
+  var path = Path.join(tmpDir, prefix); //path to a new tmp dir  
+  
+  // check if root temporary directory '/tmp' already exists
+  fs.stat(tmpDir, function(error, stats) {
+    if (!error) {
+      if (stats.isDirectory()) {
+        tmpDirExists = true;
+      }
+    } 
+  });
+
+  //try to create /tmp in root
+  if (!tmpDirExists) {
+    make_directory(context, tmpDir, function(error) {
+      if (error) {
+        callback(error, path)
+      } else {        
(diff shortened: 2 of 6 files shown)


src/filesystem/implementation.js
```diff
@@ -1678,6 +1678,31 @@ function mkdir(fs, context, path, mode, callback) {
   make_directory(context, path, callback);
 }
 
+function mkdtemp(fs, context, prefix, options, callback) { 
+  callback = arguments[arguments.length - 1];
+
+  // this function is ised to generate a random character set to append
```
bblarney: 'ised' => 'used'

AlexeiGitH: Fixed, thanks. 🙂

humphd:
@Michael-Overall fascinating.  Looking at the blame info for that file, I see the `encoding` option parsing was added in [this commit](https://github.com/nodejs/node/commit/060e5f0c0064e578c2150f13e3f91ac15fdeed92).  It seems to deal with the need to sometimes parse `Buffer`s into `String`s for filenames.  In Filer, we'll always do `utf8` encoding for `String`, so this is moot.  I would suggest allowing the `options` object to get passed in on the call, but then you can either ignore it, or just always set to `utf8`.  I don't know if we even support passing a path as a `Buffer`, or if we would want to?

AlexeiGitH:
@Michael-Overall @humphd  
Originally I added options parameter to match Node's mkdtemp, but while working on implementation I noticed that filer's make_directory function used to create directories has only path, callback parameters, so I decided to do nothing with options argument in mkdtemp.

humphd:
That's a bug in `make_directory`, we should be allowing the `options` and `mode` to come through.

Michael-Overall:
Right... Unless I'm mistaken, we usually use a Buffer to hold data for actual files, not for simple names of directories...

I'm not sure mkdir() does anything with the mode argument except to validate and convert(mask) it for use elsewhere:
see: [validateAndMask](https://github.com/filerjs/filer/blob/353290a08f8b3b15e54973fc77714324fce1519f/src/filesystem/implementation.js#L1915)

also see: [mkdir()](https://github.com/filerjs/filer/blob/353290a08f8b3b15e54973fc77714324fce1519f/src/filesystem/implementation.js#L1668)

However, this lack of using the mode argument in make_directory() may be a conscious decision based on our and Node.js' documentation. The [Node.js documentation](https://nodejs.org/api/fs.html#fs_fs_mkdir_path_mode_callback) for mkdir shows mode as being used only on Unix systems to set directory permissions, and our own Readme.md file for Filer states that "No permissions (e.g., no chown(), chmod(), etc.)." are used in Filer. Filer's ReadMe documentation for mkdir() also states that " Filer allows for, but ignores the optional mode argument used in node.js.". In other words, all we do is validate mode, and then happily ignore it...

Michael-Overall:
@AlexeiGitH @humphd 
I'd talked to @humphd  in class today and he described a case to me where file permissions in Filer were used in a project to interface a,n emulated (in a browser?) Unix VM that was run such that it needed to be sent files with permissions via Filer. 

The operation/implementation of file permissions in relation to the IndexedDB and other providers might be something to hash out so this can be made to work -- perhaps its own issue? At any rate I understood from that exchange that directory permissions were something people would use if they were implemented.

README.md
```diff
@@ -673,6 +674,30 @@ fs.mkdir('/home', function(err) {
 });
 ```
 
+#### fs.mkdtemp(prefix, options, callback)<a name="mkdtemp"></a>
+
+
```
humphd: Remove this extra blank line please.

README.md
```diff
@@ -673,6 +674,30 @@ fs.mkdir('/home', function(err) {
...
+Makes a temporary directory with prefix supplied in `path` argument. Method will append the six randomly selected characters directly to the prefix.
```
humphd: Change this to "...will append six random characters..."

README.md
```diff
@@ -306,6 +306,7 @@ var fs = new Filer.FileSystem();
 * [fs.mknod(path, mode, callback)](#mknod)
 * [fs.rmdir(path, callback)](#rmdir)
 * [fs.mkdir(path, [mode], callback)](#mkdir)
+* [fs.mkdtemp(path, [mode], callback)](#mkdtemp)
```
humphd: Here you use `[mode]` but below you refer to `options`.  Can you fix one or the other?

README.md
```diff
@@ -673,6 +674,30 @@ fs.mkdir('/home', function(err) {
 });
 ```
 
+#### fs.mkdtemp(prefix, options, callback)<a name="mkdtemp"></a>
+
+
+Makes a temporary directory with prefix supplied in `path` argument. Method will append the six randomly selected characters directly to the prefix.
+
+Asynchronous. Callback gets `(error, path)`, where path is a path to created directory.
```
humphd: should be "where `path` is the path to the created directory"

README.md
```diff
@@ -673,6 +674,30 @@ fs.mkdir('/home', function(err) {
...
+
+NOTE: Filer allows for, but ignores the optional `options` argument used in node.js.
+
+Example:
+
+```javascript
+// Create tmp directory with prefix foo
+fs.mkdtemp("foo-", function (error, path) {
```
humphd: We only support absolute paths in Filer, so this should probably be `"/foo-"`

README.md
```diff
@@ -673,6 +674,30 @@ fs.mkdir('/home', function(err) {
...
+    //A new folder foo-xxxxxx will be created. Path contains a path to created folder.    
+});
+
+fs.mkdtemp("/myDir/tmp", function (error, path) {
+    //Will create a new folder tmpxxxxxx inside myDir directory. 
+    //Will throw error if myDir does not exist    
+});
+
```
humphd: Remove this empty line.

src/filesystem/implementation.js
```diff
@@ -1678,6 +1678,31 @@ function mkdir(fs, context, path, mode, callback) {
   make_directory(context, path, callback);
 }
 
+function mkdtemp(fs, context, prefix, options, callback) { 
+  callback = arguments[arguments.length - 1];
+
+  // this function is used to generate a random character set to append 
+  // to tmp dir name to make sure it is unique
+  function generateRandom() {
```
humphd: Let's not duplicate this code.  Instead, let's modify the original code so we can reuse it.  In `shared.js`, let's do this:

```js
function generateRandom(template) {
  return template.replace(/[xy]/g, function(c) {
    var r = Math.random()*16|0, v = c == 'x' ? r : (r&0x3|0x8);
    return v.toString(16);
  });
}

function guid() {
  return generateRandom('xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx').toUpperCase();
}

/**
 * Generate a string of n random characters.  Defaults to n=6.
 */ 
function randomChars(n) {
  n = n || 6;
  var template = 'x'.repeat(n);
  return generateRandom(template);
}

...

module.exports = {
  randomChars: randomChars,
  ...
}
```

humphd: NOTE: I removed `toUpperCase()` from `generateRandom`'s code, and moved it only to `guid`.

src/filesystem/implementation.js
```diff
@@ -1678,6 +1678,31 @@ function mkdir(fs, context, path, mode, callback) {
...
+    return 'xxxxxx'.replace(/[xy]/g, function(c) {
+      var r = Math.random()*16|0, v = c == 'x' ? r : (r&0x3|0x8);
+      return v.toString(16);
+    });
+  }
+  if (prefix) {
+    var path = '/'+prefix+generateRandom();
```
humphd: Use `path.join()` for this.  Also, you can call into `shared.js` now and use `randomChars()`

src/filesystem/implementation.js
```diff
@@ -1678,6 +1678,31 @@ function mkdir(fs, context, path, mode, callback) {
...
+    make_directory(context, path, function(error) {
+      if (error) {
+        callback(error, path);
+      } else {
+        callback(null, path);
+      }
+    });
+  } else {
+    callback(new Error('filename prefix is required'), prefix);
```
humphd: Move this case up, and do it like so:

```js
if(!prefix) {
  return callback(new Error('filename prefix is required'), prefix);
}
// now deal with creating the dir...

File: src/filesystem/implementation.js [hunk 11]
```diff
@@ -1678,6 +1678,25 @@ function mkdir(fs, context, path, mode, callback) {
   make_directory(context, path, callback);
 }
 
+function mkdtemp(fs, context, prefix, options, callback) { 
+  callback = arguments[arguments.length - 1];
+  let shared = require('../shared.js');
```
humphd: Let's move this to the top of the file vs. doing it every time we call this function.  Also, use `var`, even though `let` is correct, as we haven't converted over to `const` and `let` yet.

src/filesystem/implementation.js
```diff
@@ -1678,6 +1678,25 @@ function mkdir(fs, context, path, mode, callback) {
...
+  if(!prefix) {
+    return callback(new Error('filename prefix is required'), prefix);
+  } else {
```
humphd: No need for an `else` after a `return, you can remove this.

src/filesystem/implementation.js
```diff
@@ -1678,6 +1678,25 @@ function mkdir(fs, context, path, mode, callback) {
...
+    let random = shared.randomChars(6);
+    var path = prefix+'-'+random;
```
humphd: Add some whitespace around the `+`s please.

src/filesystem/implementation.js
```diff
@@ -1678,6 +1678,25 @@ function mkdir(fs, context, path, mode, callback) {
...
 
+function mkdtemp(fs, context, prefix, options, callback) { 
+  callback = arguments[arguments.length - 1];
+  let shared = require('../shared.js');
+  if(!prefix) {
+    return callback(new Error('filename prefix is required'), prefix);
+  } else {
+    let random = shared.randomChars(6);
+    var path = prefix+'-'+random; 
+    normalize(prefix);
```
humphd: This should probably be `path = normalize(path);`

src/filesystem/implementation.js
```diff
@@ -1678,6 +1678,25 @@ function mkdir(fs, context, path, mode, callback) {
...
+    make_directory(context, path, function(error) {
+      if (error) {
```
humphd: Simplify this to: `callback(error, path);` (it will be `undefined` if not set, so you don't need to bother with the `if` checks.

src/filesystem/implementation.js
```diff
@@ -2384,6 +2403,7 @@ module.exports = {
   close: close,
   mknod: mknod,
   mkdir: mkdir,
+  mkdtemp: mkdtemp,
```
humphd: We need this to happen in https://github.com/filerjs/filer/blob/fe41a429dd3a5a57caa0df38630e53427092832b/src/filesystem/interface.js#L267-L306 too.

src/shared.js
```diff
@@ -1,8 +1,21 @@
-function guid() {
-  return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, function(c) {
+function generateRandom(template) {
```
humphd: There's a rebase necessary for this file, can you update your branch and fix the conflict?

tests/spec/fs.mkdtemp.spec.js
```diff
@@ -0,0 +1,50 @@
...
+
+  it('should craete temp dir with specified prefix', function(done) {
+    var fs = util.fs();
+    fs.mkdtemp('/foo', function(error, path) {
+      expect(error).not.to.exist;
+      expect(path).to.match(/foo.{6}/);
+      done();
+    });
+  });
+  it('should craete temp dir inside existing directory', function(done) {
```
humphd: Newline above this please

tests/spec/fs.mkdtemp.spec.js
```diff
@@ -0,0 +1,50 @@
...
+  it('should be a function', function() {
+    var fs = util.fs();
+    expect(fs.mkdtemp).to.be.a('function');
+  });
+
+  it('should craete temp dir with specified prefix', function(done) {
+    var fs = util.fs();
+    fs.mkdtemp('/foo', function(error, path) {
+      expect(error).not.to.exist;
+      expect(path).to.match(/foo.{6}/);
```
humphd: I think you want to use `/^\w{7}$/` instead, or perhaps ``/^\-\w{6}$/`.  Same for the others in this file.

tests/spec/fs.mkdtemp.spec.js
```diff
@@ -0,0 +1,50 @@
...
+        expect(path).to.match(/foo.{6}/);
+        done();
+      });
+    });
+  });
+
+  it('should not create temp dir without prefix', function(done) {
+    var fs = util.fs();
+    fs.mkdtemp('', function(error, path) {
+      expect(error).to.exist;
```
humphd: Should proably check for an `EINVAL` here, similar to what happens below ` expect(error.code).to.be.equal('ENOENT');`

tests/spec/fs.mkdtemp.spec.js
```diff
@@ -0,0 +1,50 @@
...
+      expect(path).to.be.equal('');
+      done();
+    });
+  });
+  it('should not create temp dir inside non existing dir', function(done) {
```
humphd: Newline above this.

tests/spec/fs.mkdtemp.spec.js
```diff
@@ -0,0 +1,50 @@
...
+    var fs = util.fs();
+    fs.mkdtemp('/doesNotExists/foo', function(error, path) {
+      expect(error).to.exist;
+      expect(error.code).to.be.equal('ENOENT');
+      expect(path).to.exist;
+      done();
+    });
+  });
+});
```
humphd: Newline at end of file too please.

tests/spec/fs.mkdtemp.spec.js
humphd: This got missed throughout this file, can you update your regex?  It's working now, but only testing the first 6 characters of 7, and using `.` is too much I think.

humphd:
OK, I've rebased this, fixed a few things that were still outstanding, updated a test that was failing, and merged this into `master` as https://github.com/filerjs/filer/commit/0eb7066a861d9d4506a4b2b2bb10865379aab82c.  Great work!

PR #448, opened 2018-09-22
Fix #447: Added test to append numbers in fs.appendFile function

### LINKED ISSUE #447: Add test case for fs.appendFile
I noticed there was no test case for appending non-string numbers to a file

I would like to add a test case for:
-appending numbers to a file in the fs.appendFile function.

PopeSpaceous:
Hi, here is a test for fs.appendFile to append numbers to a file. 
Let me know if there are any changes you would like made.

### DIFF
--- a/tests/spec/fs.appendFile.spec.js
+++ b/tests/spec/fs.appendFile.spec.js
@@ -124,4 +124,20 @@ describe('fs.appendFile', function() {
       });
     });
   });
+  
+  it('should accept numbers and appends them to the file', function(done) {
+    var fs = util.fs();
+    var contents = 'This is a file.';
+    var more = 10000;
+
+    fs.appendFile('/myfile', more, 'utf8', function(error) {
+      if(error) throw error;
+
+      fs.readFile('/myfile', 'utf8', function(error, data) {
+        expect(error).not.to.exist;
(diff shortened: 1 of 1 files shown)


tests/spec/fs.appendFile.spec.js
```diff
@@ -124,4 +124,20 @@ describe('fs.appendFile', function() {
...
+    var fs = util.fs();
+    var contents = 'This is a file.';
+    var more = 10000;
+
+    fs.appendFile('/myfile', more, 'utf8', function(error) {
+      if(error) throw error;
+
+      fs.readFile('/myfile', 'utf8', function(error, data) {
+        expect(error).not.to.exist;
+        expect(data).to.equal(contents + more);
```
humphd: This doesn't make sense: you are expecting to have `'This is a file.10000'`, but you never write `contents` to disk.  You'd either need to add a `writeFile` above that does this first (before your `appendFile`, or just get rid of `contents` completely from this test and test for the contents to be `'10000'`.

humphd:
Finished this in https://github.com/filerjs/filer/commit/1c450ae8e758e29d27016c7a18893f192a97b2fb and rebased, landed.

PR #449, opened 2018-09-22
Fixed #408: The error checking statements were using the wrong variable names in the fs.read() example in doc. 

### LINKED ISSUE #408: Error in doc's for fs.read
In the error checking statements the wrong variable names were used.

```js
    if(err) throw error;
```
should be
```js
    if(err) throw err;
```
When you run this code you get the following error

```console
  if(err) throw error;
          ^

ReferenceError: error is not defined
```

Link to Doc [Here](https://github.com/filerjs/filer#fsreadfd-buffer-offset-length-position-callback)

Below is the code from the docs
```js
// Create a file with the following bytes.
var buffer = new Filer.Buffer([1, 2, 3, 4, 5, 6, 7, 8]);

fs.open('/myfile', 'w', function(err, fd) {
  if(err) throw error; // Error here should be if(err) throw err;

  var expected = buffer.length, written = 0;
  function writeBytes(offset, position, length) {
    length = length || buffer.length - written;

    fs.write(fd, buffer, offset, length, position, function(err, nbytes) {
      if(err) throw error; // Error here should be if(err) throw err;

      // nbytes is now the number of bytes written, between 0 and buffer.length.
      // See if we still have more bytes to write.
      written += nbytes;

      if(written < expected)
        writeBytes(written, null);
      else
        fs.close(fd);
    });
  }

  writeBytes(0, 0);
});
````

coreyjjames:
This pull request fixes the typo in the error checking statements in the docs for fs.read().

### DIFF
--- a/README.md
+++ b/README.md
@@ -908,11 +908,11 @@ Example:
 
 ```javascript
 fs.open('/myfile', 'r', function(err, fd) {
-  if(err) throw error;
+  if(err) throw err;
 
   // Determine size of file
   fs.fstat(fd, function(err, stats) {
-    if(err) throw error;
+    if(err) throw err;
 
     // Create a buffer large enough to hold the file's contents
     var nbytes = expected = stats.size;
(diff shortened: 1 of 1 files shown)


SeanPrashad:
🚢 it!!

humphd:
This should be shipped in the latest version.

PR #452, opened 2018-09-22
Fix issue-426: Add  the test for fsPromises.readdir when the path exsts, and is a directory

ywpark1:
I added two tests for fsPromises.readdir function to ensure that fsPromises.readdir is a function, and it returns the contents of the directory if the path given exists and is a directory.

I was not sure if I should create new group for fsPromises.readdir test cases(i.e. describe(...) ), so I simply added under fs.readdir test cases by adding '(promise)' in the front.

I also checked [PR-issue-400](https://github.com/filerjs/filer/pull/409). It seems like there is a conflict between my tests and @jespirit 's test in terms of how fsPromises tests are written.

Any advice for this?

### DIFF
--- a/tests/spec/fs.readdir.spec.js
+++ b/tests/spec/fs.readdir.spec.js
@@ -54,4 +54,33 @@ describe('fs.readdir', function() {
       });
     });
   });
+
+  it('(promise) should be a function', function() {
+    var fsPromises = util.fs().promises;
+    expect(fsPromises.readdir).to.be.a('function');
+  });
+  
+  it('(promise) should return a list of files from an existing directory', function() {
+    var fsPromises = util.fs().promises;
+
+    return fsPromises.mkdir('/tmp')
+      .then(() => {
+        return fsPromises.stat('/');
+      })
+      .then(stats => {
+        expect(stats).to.exist;
+        expect(stats.isDirectory()).to.be.true;
+      })
+      .then(() => {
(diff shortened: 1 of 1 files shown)


jespirit:
I think you should separate the tests in two different sections because the functions are different in terms of their arguments and what they return, so `fs.readdir` is quite different than `fs.promises.readdir`.

I agree there seems to be a syntax conflict for formatting the tests and we should try agree on a consensus so that the code looks like it was written by one person. You can refer to my comment I made here (at the bottom) regarding the syntax about promises in another pull request https://github.com/filerjs/filer/pull/402#issuecomment-423194993

For example, I could rewrite your tests to look like mine, or vice-versa.

```javascript
.then(() => fsPromises.readdir('/'))
// same as
.then(() => return fsPromises.readdir('/'))
// same as
.then(() => {
  return fsPromises.readdir('/');
})
// same as
.then(function() {
  return fsPromises.readdir('/');
})
```

tests/spec/fs.readdir.spec.js
```diff
@@ -54,4 +54,33 @@ describe('fs.readdir', function() {
       });
     });
   });
+
+  it('(promise) should be a function', function() {
```
YuechengWu: Just to nitpick here, you can use () => {...} to declare a function instead of function() {...}. It's a shorter syntax so less typing.

PR #454, opened 2018-09-23
Fix#453: Added fsPromises.link() tests

Dragomegak:
Implemented the promise tests of link to the fs.link.spec.js based on the non promised version

### DIFF
--- a/tests/spec/fs.link.spec.js
+++ b/tests/spec/fs.link.spec.js
@@ -134,12 +134,11 @@ describe('fs.promises.link', function() {
       fs.close(fd, function(error) {
         if(error) throw error;
 
-        fs.promises.link('/myfile', '/myotherfile', function(error) {
-          if(error) throw error;
+        fs.promises.link('/myfile', '/myotherfile').then(function(){
 
           fs.stat('/myfile', function(error, result) {
             if(error) throw error;
-
+      
             var _oldstats = result;
             fs.stat('/myotherfile', function(error, result) {
               expect(error).not.to.exist;
@@ -150,7 +149,9 @@ describe('fs.promises.link', function() {
               expect(result.type).to.equal(_oldstats.type);
               done();
             });
-          });
+          },
+          function(error){throw error;}
+          );
         });
       });
     });
@@ -190,11 +191,10 @@ describe('fs.promises.link', function() {
       fs.symlink('/', '/myfileLink', function (error) {
         if (error) throw error;
 
-        fs.promises.link('/myfileLink', '/myotherfile', function (error) {
+        fs.promises.link('/myfileLink', '/myotherfile').then(function(){
           if (error) throw error;
-
+      
           fs.lstat('/myfileLink', function (error, result) {
-            if (error) throw error;
             
             var _linkstats = result;
             fs.lstat('/myotherfile', function (error, result) {
@@ -206,7 +206,9 @@ describe('fs.promises.link', function() {
               expect(result.nlinks).to.equal(2);
               done();
             });
-          });
+          },
+          function(error){throw error;}
+          );
         });
       });
     });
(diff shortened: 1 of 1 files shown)


tests/spec/fs.link.spec.js
```diff
@@ -111,3 +111,119 @@ describe('fs.link', function() {
...
+  it('should create a link to an existing file', function(done) {
+    var fs = util.fs();
+
+    fs.open('/myfile', 'w+', function(error, fd) {
+      if(error) throw error;
+
+      fs.close(fd, function(error) {
+        if(error) throw error;
+
+        fs.promises.link('/myfile', '/myotherfile', function(error) {
```
dnguneratne: Should follow the promise syntax
```js
fs.promises.link('/myfile', '/myotherfile')
	.then()..
```

tests/spec/fs.link.spec.js
```diff
@@ -111,3 +111,119 @@ describe('fs.link', function() {
...
+
+
+
+
+describe('fs.promises.link', function() {
+  beforeEach(util.setup);
+  afterEach(util.cleanup);
+
+  it('should be a function', function() {
+    var fs = util.fs();
```
dnguneratne: You can use the promise API directly with:
`var fs = util.fs().promises;`

tests/spec/fs.link.spec.js
Dragomegak: Done for tests:  
- create a link to an existing file
- not follow symbolic links 

although I could not get it running for not allowing links to a directory

tests/spec/fs.link.spec.js
Dragomegak: I tried checking out other implementations of this so I could see how I could do this but other people's code did not do this. By chance do you mind giving me more information about how to implement it?

humphd: I think it's fine either way.  The `util.fs().promises` way would be good if you were needing to use it multiple times, and you could cache it like this.  In this case, it's just once, so either or is fine.

tests/spec/fs.link.spec.js
```diff
@@ -111,3 +111,121 @@ describe('fs.link', function() {
...
+  });
+
+  it('should create hard link to identical data node', function(done) {
+    var fs = util.fs();
+    var contents = 'Hello World!';
+
+    fs.writeFile('/file', contents, function(err) {
+      if(err) throw err;
+
+      fs.promises.link('/file', '/hlink', function(err) {
```
humphd: This is mixing callbacks and promises.  You shouldn't be passing a callback to link here.

tests/spec/fs.link.spec.js
```diff
@@ -111,3 +111,88 @@ describe('fs.link', function() {
...
+
+  it('should be a function', function() {
+    var fs = util.fs();
+    expect(fs.promises.link).to.be.a('function');
+  });
+
+  it('should create a link to an existing file', function(done) {
+    var fs = util.fs();
+
+    fs.open('/myfile', 'w+', function(error, fd) {
```
humphd: You're mixing callbacks and promises in the same test, which is a bit confusing.  Here's how you should write something like this:

```js
var oldStats;

return fsPromises.open('/myfile', 'w+')
  .then(fd => fsPromises.close(fd))
  .then(() => fsPromises.stat('/myfile'))
  .then(result => {
    oldStats = result;
    return fsPromises.stat('/myotherfile')
  })
  .then(result => {
    // do all your expect assertions here)
  });
```

Can you fix all your tests so they only use promises?

humphd:
Fixed, rebased, and landed in https://github.com/filerjs/filer/commit/16e6b3e1c9ba9840dba0815ef3af1c9d459e1897.

PR #455, opened 2018-09-23
Fix #419: Add test for fsPromises.open() when the path does not exist

### LINKED ISSUE #419: Add test for fsPromises.open() when the path does not exist
The following could be used with fsPromises.open() to test when the path given does not exist.

```js
it('should return an error if the parent path does not exist', function(done) {
    var fs = util.fs();

    fs.open('/tmp/myfile', 'w+', function(error, result) {
      expect(error).to.exist;
      expect(result).not.to.exist;
      done();
    });
  });
```

I'd like to work on this issue and add the change to the file if it's okay.

struong9:
Added a test for fsPromises.open() when the path doesn't exist

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,9 +14,7 @@
   "version": "0.0.44",
   "author": "Alan K <ack@modeswitch.org> (http://blog.modeswitch.org)",
   "homepage": "http://filerjs.github.io/filer",
-  "bugs": {
-    "url": "https://github.com/filerjs/filer/issues"
-  },
+  "bugs": "https://github.com/filerjs/filer/issues",
   "license": "BSD-2-Clause",
   "scripts": {
     "eslint": "npm run lint",
@@ -33,7 +31,7 @@
   },
   "repository": {
     "type": "git",
-    "url": "git+https://github.com/filerjs/filer.git"
+    "url": "https://github.com/filerjs/filer.git"
   },
   "dependencies": {
     "base64-arraybuffer": "^0.1.5",
@@ -52,9 +50,5 @@
     "parcel-bundler": "^1.9.6"
   },
   "main": "./src/index.js",
-  "browser": "./dist/filer.min.js",
-  "directories": {
-    "lib": "lib",
-    "test": "tests"
-  }
+  "browser": "./dist/filer.min.js"
 }
--- a/package.json
+++ b/package.json
@@ -14,7 +14,9 @@
   "version": "0.0.44",
   "author": "Alan K <ack@modeswitch.org> (http://blog.modeswitch.org)",
   "homepage": "http://filerjs.github.io/filer",
-  "bugs": "https://github.com/filerjs/filer/issues",
+  "bugs": {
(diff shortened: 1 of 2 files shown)


tests/spec/fs.open.spec.js
```diff
@@ -137,3 +137,23 @@ describe('fs.open', function() {
...
 });
+
+/**
+ * fsPromise.open() Tests
+ **/
+describe('fsPromises.open', function() {
+  beforeEach(util.setup);
+  afterEach(util.cleanup);
+
+  it('should return an error if the parent path does not exist', function(done) {
```
Dragomegak: Good job with the tests, I don't see anything wrong.

package.json
```diff
@@ -14,7 +14,9 @@
   "version": "0.0.44",
   "author": "Alan K <ack@modeswitch.org> (http://blog.modeswitch.org)",
   "homepage": "http://filerjs.github.io/filer",
-  "bugs": "https://github.com/filerjs/filer/issues",
+  "bugs": {
```
Dragomegak: Although you added the package.json, I'm guessing it was pushed when you added everything and pushed it, remember to use the command git status to see the files that were changed and use git add filepath/file to add it singularily.

struong9: Sorry I'm not familiar with Git, but by this do you mean I should only push <b>fs.open.spec.js</b> and discard changes for <b>package.json</b>?

humphd: Exactly @struong9, you can revert this change to `package.json` with `git checkout master package.json` and then `git add` and `git commit` and `git push`

tests/spec/fs.open.spec.js
ooHAoo: You can get rid of 'done' argument as it's not needed with promises

tests/spec/fs.open.spec.js
```diff
@@ -137,3 +137,23 @@ describe('fs.open', function() {
...
+    var fsPromises = util.fs().promises;
+
+    fsPromises.open('/tmp/myfile', 'w+')
+      .then(result => expect(result).not.to.exist)
+      .catch(error => {
+        expect(error).to.exist;
+        expect(error.code).to.equal('ENOENT');
+      })
+      .then(() => done());
```
ooHAoo: this line is not needed (.then(() => done()); since we are using promises :)

tests/spec/fs.open.spec.js
```diff
@@ -137,3 +137,23 @@ describe('fs.open', function() {
...
+ * fsPromise.open() Tests
+ **/
+describe('fsPromises.open', function() {
+  beforeEach(util.setup);
+  afterEach(util.cleanup);
+
+  it('should return an error if the parent path does not exist', function(done) {
+    var fsPromises = util.fs().promises;
+
+    fsPromises.open('/tmp/myfile', 'w+')
```
ooHAoo: Please return this function -> fsPromises(....)

struong9:
@humphd Added.

humphd:
Thanks, merged.

Issue #427, opened 2018-09-21
Add tests for fsPromises.rmdir


Casva:
There are currently no tests for [fsPromises.rmdir(path)](https://nodejs.org/docs/latest/api/fs.html#fs_fspromises_rmdir_path)

I would like to create the tests for [when the provided path doesn't exist](https://github.com/filerjs/filer/blob/master/tests/spec/fs.rmdir.spec.js#L13) and for [when the user tries to remove the root directory](https://github.com/filerjs/filer/blob/master/tests/spec/fs.rmdir.spec.js#L23)

Could I get this issued assigned to me?
Thanks!

humphd:
Go for it @Casva, it's all yours.

PR #457, opened 2018-09-23
Issue 427 - Add tests to fsPromises.rmdir when directory doesn't exist and when trying to delete root directory

Casva:
For issue #427, I've added two tests for fsPromises.rmdir for the following:

1) The user tries to remove a non-existent directory
2) The user tries to remove the root directory

Hopefully everything is fine!

### DIFF
--- a/tests/spec/fs.rmdir.spec.js
+++ b/tests/spec/fs.rmdir.spec.js
@@ -97,4 +97,28 @@ describe('fs.rmdir', function() {
       });
     });
   });
+
+  it('(promise) should be a function', function() {
+    var fsPromises = util.fs().promises;
+    expect(fsPromises.rmdir).to.be.a('function');
+  });
+  
+  it('(promise) should return an error if the path does not exist', function() {
+    var fsPromises = util.fs().promises;
+      return fsPromises.rmdir('/tmp/mydir')
+        .catch(error => {
+          expect(error).to.exist;
+          expect(error.code).to.equal('ENOENT');
+        });
+  });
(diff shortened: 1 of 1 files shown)


tests/spec/fs.rmdir.spec.js
```diff
@@ -97,4 +97,28 @@ describe('fs.rmdir', function() {
...
+  it('(promise) should return an error if attempting to remove the root directory', function() {
+    var fsPromises = util.fs().promises;
+    return fsPromises.rmdir('/')
+      .catch(error => {
+        expect(error).to.exist;
+        expect(error.code).to.equal('EBUSY');
+      });
+  });
 });
+
```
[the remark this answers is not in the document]
Casva: Ah! I must have missed that thanks! 
Will change it ASAP

tests/spec/fs.rmdir.spec.js
```diff
@@ -97,4 +97,27 @@ describe('fs.rmdir', function() {
       });
     });
   });
+
+  it('(promise) should be a function', function() {
```
YuechengWu: Just to nitpick here, you can use () => {...} instead of function () {...}. It's a shorter syntax.

biskit1: @YuechengWu other PR comments on this syntax suggestion seem to recommend maintaining consistent syntax throughout the file; because the other (non-promise) cases are not using the arrow syntax, it might be best to leave it as is.

PR #458, opened 2018-09-23
Fix #405 - Added tests for chmod and fchmod using Promises

### LINKED ISSUE #405: Add a test for fsPromises chmod
There is currently no test for the promises version of chmod in filer/tests/spec/fs.chmod.spec.js
I can add a test case for this. 

I can also look at fchmod but I'm unsure if there's a promises version. chmod is documented here https://nodejs.org/api/fs.html#fs_fspromises_chmod_path_mode but there is no fchmod promises documentation. I can test it out when writing the chmod test case and if fchmod also works with promises I could include it in my PR.

danielbogomazov:
### Added test cases for chmod and fchmod using Promises

Fix issue #405 

## Tests:

- [x] Whether they are functions
- [x] Whether they return a Promise object
- [x] Whether they change the mode of a file

**Note**
I separated the chmod and fchmod tests to pinpoint any possible future problems. If we use the non-promise test model, it will only report on fchmod being broken - even if both functions break (since we test for fchmod before chmod). We should consider splitting up the non-promises tests too. Maybe put fchmod into its own file?

Also, should the mkdir test case be moved to the [mkdir test file](https://github.com/filerjs/filer/blob/master/tests/spec/fs.mkdir.spec.js)? 
Even though the main point of the test is to check whether the correct mode is being set by the function, it makes more sense to me to place a mkdir test in a mkdir test file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ bower_components
 *~
 dist/filer-issue225.js
 .vscode
+.DS_Store
 
 # Parcel build dirs
 .cache
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,6 @@ bower_components
 *~
 dist/filer-issue225.js
 .vscode
-.DS_Store
 
 # Parcel build dirs
 .cache
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
   },
   "devDependencies": {
     "chai": "^4.1.2",
-    "eslint": "^5.0.1",
+    "eslint": "^5.6.0",
     "karma": "^3.0.0",
     "karma-chai": "^0.1.0",
     "karma-chrome-launcher": "^2.2.0",
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
   },
   "devDependencies": {
     "chai": "^4.1.2",
-    "eslint": "^5.6.0",
+    "eslint": "^5.0.1",
     "karma": "^3.0.0",
     "karma-chai": "^0.1.0",
     "karma-chrome-launcher": "^2.2.0",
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -4,6 +4,7 @@ bower_components
 *~
 dist/filer-issue225.js
 .vscode
+.DS_Store
```
coreyjjames: Here

package.json
```diff
@@ -40,7 +40,7 @@
   },
   "devDependencies": {
     "chai": "^4.1.2",
-    "eslint": "^5.0.1",
+    "eslint": "^5.6.0",
```
coreyjjames: Here

danielbogomazov:
@coreyjjames 
Thanks for the heads-up. The files have been removed from the PR.

tests/spec/fs.chmod.spec.js
```diff
@@ -71,3 +71,71 @@ describe('fs.chmod, fs.fchmod', function() {
...
+
+  it('should return a promise', function() {
+    var fsPromise = util.fs().promises;
+    expect(fsPromise.chmod()).to.be.a('Promise');
+  });
+
+  it('should allow for updating mode of a given file', function() {
+    var fsPromise = util.fs().promises;
+
+    return fsPromise.open('/file', 'w')
```
biskit1: I'm wondering about how you've nested these `.then()` and `.catch()`... 
Is there a better way these can be nested? 
Maybe instead of doing the `return` within `return`, just call the functions and have the `.then()` and `.catch()` for that function nested separately? 
i.e
``` 
var fsPromise = util.fs().promises;
     return fsPromise.open('/file', 'w')
         .then(()=>{
              fsPromise.chmod('/file', 0o444)
                  .then(...)
                  .catch(...);
          }) 
         .catch(...);
```
I'm not sure what's considered best practice, or if it there is even a 'better' or 'worse' way to do it at all...

danielbogomazov: I've removed the redundant ``.catch()`` blocks. 
Thanks for the review.

Issue #417, opened 2018-09-20
Add test for fs.promises.rename() to test for renaming an existing directory


bblarney:
I noticed that [a test has been added](https://github.com/filerjs/filer/pull/416) for promises.rename() to test for renaming an existing file.

I would like to create a test for renaming an existing directory using promises, and add it to the [test file](https://github.com/filerjs/filer/blob/master/tests/spec/fs.rename.spec.js)

humphd:
Go for it @bblarney.

PR #460, opened 2018-09-23
Issue 417 - Added test for fs.promises.rename to rename existing directory

bblarney:
Added test for fs.promises.rename() to rename a directory which already exists.

[Link to issue](https://github.com/filerjs/filer/issues/417)

Any feedback appreciated!

### DIFF
--- a/tests/spec/fs.rename.spec.js
+++ b/tests/spec/fs.rename.spec.js
@@ -75,8 +75,7 @@ describe('fs.rename', function() {
 
     return fsPromises.mkdir('/mydir')
       .then(() => {
-        return fsPromises.rename('/mydir', '/myotherdir')
-          .catch((error) => expect(error).not.to.exist);
+        return fsPromises.rename('/mydir', '/myotherdir');
       })
       .then(() => {
         return fsPromises.stat('/mydir')
--- a/tests/spec/fs.rename.spec.js
+++ b/tests/spec/fs.rename.spec.js
@@ -74,22 +74,17 @@ describe('fs.rename', function() {
     var fsPromises = util.fs().promises;
 
     return fsPromises.mkdir('/mydir')
-      .then(() => {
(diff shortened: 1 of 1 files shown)


tests/spec/fs.rename.spec.js
```diff
@@ -70,6 +70,33 @@ describe('fs.rename', function() {
     });
   });
 
+  it('should rename an existing directory (using promises)', () => {
+    var fsPromises = util.fs().promises;
+
+    return fsPromises.mkdir('/mydir')
+      .then(() => {
+        return fsPromises.rename('/mydir', '/myotherdir')
+          .catch((error) => expect(error).not.to.exist);
```
Casva: Logical error here: By catching an error, you would expect an error to already exist
You could just remove this line

tests/spec/fs.rename.spec.js
```diff
@@ -70,6 +70,33 @@ describe('fs.rename', function() {
...
+        return fsPromises.stat('/mydir')
+          .catch((error) => {
+            expect(error).to.exist;
+            expect(error.code).to.equal('ENOENT');
+          });
+      })
+      .then(() => {
+        return fsPromises.stat('/myotherdir')
+          .catch((error, result) => {
+            expect(error).not.to.exist;
```
Casva: Same as above: you would expect an error to exist if you are catching it
You probably want something along these lines:

```js
.then(() => fsPromises.stat('/myotherdir'))
.then((error, result) => {
     expect(error).not.to.exist;
     expect(result.nlinks).to.equal(1);
})

Comment (humphd) [reply tests/spec/fs.rename.spec.js]: Instead of doing `.catch((error, result) => {` you can do:

```js
.then(result => {
  expect(result.nlinks).to.equal(1);
}
```

bblarney:
@humphd I've added some new changes, ready to be reviewed

humphd:
Thanks @bblarney, merged.

PR #462, opened 2018-09-23
Fix #421 - Add test for fs.promises.link to ensure it returns a Promise

### LINKED ISSUE #421: Add test for fsPromises.link to check if it returns a Promise
From what I can tell it would be useful if there was a test for the `fsPromises.link` method to check that it returns a promise object. This test would be added to `fs.link.spec.js` and would require a new describe block in the same file for said `fsPromises.link` method. It would look something like the following: 

```
it('should return a Promise object', function(done) {
   //code goes here
});
```
I would like to work on this issue if possible.

jkmdev:
Closes issue [#421](https://github.com/filerjs/filer/issues/421)

Adds test for `fs.promises.link` in `fs.link.spec.js` to test if the method returns a promise. 
If anyone has any comments/suggestions relating to this pull request, feel free to say so.

### DIFF
--- a/tests/spec/fs.link.spec.js
+++ b/tests/spec/fs.link.spec.js
@@ -1,6 +1,7 @@
 var util = require('../lib/test-utils.js');
 var expect = require('chai').expect;
 
+
 describe('fs.link', function() {
   beforeEach(util.setup);
   afterEach(util.cleanup);
@@ -80,7 +81,7 @@ describe('fs.link', function() {
 
           fs.lstat('/myfileLink', function (error, result) {
             if (error) throw error;
-            
+
             var _linkstats = result;
             fs.lstat('/myotherfile', function (error, result) {
               expect(error).not.to.exist;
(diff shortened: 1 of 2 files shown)


tests/spec/fs.link.spec.js
```diff
@@ -110,4 +109,18 @@ describe('fs.link', function() {
...
+
+});
+
+describe('fs.promises.link', function() {
+
+  beforeEach(util.setup);
+  afterEach(util.cleanup);
+
+  it('should return a promise', function() {
+    var fs = util.fs().promises;
```
danielbogomazov: Maybe consider changing the `fs` variable to `fsPromise` to differentiate it from the non-promise FileSystem?

jkmdev: Thanks for the feedback! I agreed with it and decided to change the variable's name to make it more accurate.

charBap:
Code looks good and precise!! I have been learning a lot about Promises these past few days and this request has helped me learn more about it.
Just a tiny nit pick ... Github is telling us to remove blank lines 4 and 84 and also delete line 125.

PR #464, opened 2018-09-23
Fix#422 - Installed karma-coverage and istanbul dependencies

MarkKrutik:
This was an addon that will allow to add coverage to the code. It is not fully implemented though. If you want any changes to be done, please let me know.


karma.conf.js
```diff
@@ -1,18 +1,31 @@
 module.exports = function(config) {
   config.set({
+    npm_config_coverage: ['true'],
```
humphd: I'm not familiar with this, what does it do?

humphd:
@MarkKrutik thanks for starting this.  I spent some time this week trying to get coverage to work with Karma and Parcel, and in the end, I don't think it's possible.  I've got another solution based on https://github.com/istanbuljs/nyc, which I'll do a PR for in a sec and CC you.

PR #466, opened 2018-09-23
Fix #445 test cases for fs.unwatchFile

### LINKED ISSUE #445: Add test case for fs.unwatch
There is no test case for [fs.unwatch](https://nodejs.org/api/fs.html#fs_fs_unwatchfile_filename_listener).

I would like to add test case for:
-should be a function
-calling a filename not being watched will not produce errors

VinceBackpack:
Added a new file to test the `fs.unwatchFile` function.

### DIFF
--- a/tests/index.js
+++ b/tests/index.js
@@ -37,6 +37,7 @@ require('./spec/trailing-slashes.spec');
 require('./spec/times.spec');
 require('./spec/time-flags.spec');
 require('./spec/fs.watch.spec');
+require('./spec/fs.unwatchFile.spec');
 require('./spec/errors.spec');
 require('./spec/fs.shell.spec');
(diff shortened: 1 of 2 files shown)


yatsenko-julia:
Hey, I've tried to test your code locally and your tests do not run because as @humphd noticed a new file that you've created is not included in index.js. After I added the needed line and ran the test, two failures appeared. I've checked your code - it looks great, the problem is that a method unwatchFile does not exist in filer. I think you need to contact @humphd and discuss whether or not this method should be implemented or you should work on other tests.

humphd:
Fixed this up a bit, rebased, and landed in https://github.com/filerjs/filer/commit/b0809272a125d697eeabdc1c56cae6230237264a.

PR #471, opened 2018-09-24
added tests for fs.promises.chown() closes #465

### LINKED ISSUE #465: add tests for fsPromises.chown()
currently there are no tests involving the promises version of the fs.chown() function. 

fs.chown() currently has tests for
- should be a function
- should automatically set file uid/gud to 0
- should allow updating of uid/gid

I would like to be assigned to implement these same tests for fs.promises.chown().

Thanks,

Thomas N

ThomasNolte:
added tests to check if the fs.promises.chown() function exists and if it can update the gid & uid for a file.


Mordax:
/* Side note, you should reference the issue number in the title of your PR. If you add `closes #465` (which is your issue number) it'll automatically close the issue for you when your PR is merged.  */

tests/spec/fs.chown.spec.js
```diff
@@ -62,3 +62,51 @@ describe('fs.chown, fs.fchown', function() {
...
+describe('fs.promises.chown', function(){
+  beforeEach(util.setup);
+  afterEach(util.setup);
+
+  it('should be a function', function(){
+    var fs= util.fs();
+    expect(fs.promises.chown).to.be.a('function');
+  });
+
+  it('should allow updating uid and gid for a file', function(done) {
```
humphd: You're mixing callbacks and Promises here.  Let's simplify this test a bit and just use Promises.  Here's how I'd do this (code untested, just to get you started):

```js
it('should allow updating uid and gid for a file', () => {
  
  return fsPromises.writeFile('/file', 'data')
    .then(() => fsPromises.chown('/file', 500, 500))
    .then(() => fsPromises.stat('/file'))
    .then(stats => {
        expect(stats.uid).to.equal(500);
        expect(stats.gid).to.equal(500);
    });
});
```

Here I'm not using a `done` callback.  Instead, I'm returning the Promise created by my initial `fsPromises.writeFile` call.  I think chain `.then()` calls onto this, and in each case return *another* promise to the next `.then()` in the chain until I get my `stats` and can do the assertions.  If anything fails in this chain, the test will blow up because we have nothing in a `catch`, which is what we want (i.e., fail if anything goes wrong).

Does this make sense?  Let me know if you have questions.  I haven't tested this code, so it might need some tweaks still.

PR #472, opened 2018-09-24
Added test for fs.read when attempting to read a file that does not exist

CometS1:
Created a test for fs.read attempting to read a file that does not exist to check to see if an error is received.

### DIFF
--- a/tests/spec/fs.read.spec.js
+++ b/tests/spec/fs.read.spec.js
@@ -2,28 +2,28 @@ var Filer = require('../../src');
 var util = require('../lib/test-utils.js');
 var expect = require('chai').expect;
 
-describe('fs.read', function() {
+describe('fs.read', function () {
   beforeEach(util.setup);
   afterEach(util.cleanup);
 
-  it('should be a function', function() {
+  it('should be a function', function () {
     var fs = util.fs();
     expect(fs.read).to.be.a('function');
   });
 
-  it('should read data from a file', function(done) {
+  it('should read data from a file', function (done) {
     var fs = util.fs();
     var wbuffer = new Filer.Buffer([1, 2, 3, 4, 5, 6, 7, 8]);
     var rbuffer = new Filer.Buffer(wbuffer.length);
     rbuffer.fill(0);
 
-    fs.open('/myfile', 'w+', function(error, fd) {
-      if(error) throw error;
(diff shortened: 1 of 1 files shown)


tests/spec/fs.read.spec.js
```diff
@@ -87,4 +87,17 @@ describe('fs.read', function() {
...
+
+    var fd = 0;
+
+    fs.read(fd, buffer, 0, buffer.length, undefined, function(error, result) {
+      expect(error).to.exist;
+      expect(result).not.to.exist;
+      done();
+    });
+  });
+});
```
0xazure: nit: file should end in a blank line

tests/spec/fs.read.spec.js
```diff
@@ -87,4 +87,17 @@ describe('fs.read', function() {
...
   });
-});
+
+  it('should fail to read a file that does not exist', function(done) {
+    var buffer = new Filer.Buffer(0);
+    var fs = util.fs();
+
+    var fd = 0;
+
+    fs.read(fd, buffer, 0, buffer.length, undefined, function(error, result) {
```
0xazure: instead of `position = undefined`, i would model this test case more closely after the test for [reading from a file](https://github.com/filerjs/filer/blob/fd3de6be2c1c19421d8c83e6791cd4b19afb84c7/tests/spec/fs.read.spec.js#L22). we don't want to change too many variables at a time in our tests, because each test should only test one thing. in this case, we're testing what happens when `fd` is zero _and_ `position` is `undefined`, instead of just one or the other.

tests/spec/fs.read.spec.js
```diff
@@ -87,4 +87,18 @@ describe('fs.read', function() {
...
+
+  it('should fail to read a file that does not exist', function(done) {
+    var fs = util.fs();
+
+    var fd = 0;
+    var rbuffer = new Filer.Buffer(8);
+    rbuffer.fill(0);
+
+    fs.read(fd, rbuffer, 0, rbuffer.length, 0, function(error, result) {
+      expect(error).to.exist;
```
humphd: We should also deal with the specific error code. In Filer that would (currently) be `'EBADF'`.  In note that in node, it would be: `Error: ENXIO: no such device or address, read errno: -6, code: 'ENXIO'`.  This would require a fix to our implementation if we want to match 100% (not sure it's that important).

PR #476, opened 2018-09-24
Fixes #420 - Setting a mode for a new file, when opening it. 

### LINKED ISSUE #420: Currently, the open functionality has an argument for a mode but it is not handled.
As seen here Filer's [`open` function](https://github.com/filerjs/filer/blob/fd3de6be2c1c19421d8c83e6791cd4b19afb84c7/src/filesystem/implementation.js#L1610) takes an argument for a mode. My understanding is, this mode will be used if the open function needs to create the file. (Changing a mode if a file exists seems like a security issue).

We need to handle the mode argument, and then change the `open_file` function to better support a mode argument as well. 

A test should be created, whereby you open a new file for writing, and set the mode for it. 

Edit: I'll try to tackle this.

steaward:
This fixes issue (https://github.com/filerjs/filer/issues/420) 

1.  a `mode` has been added to the [`fs.open_file`](https://github.com/filerjs/filer/blob/e3a1187ef94d5fc627768cee13457417e7ea0035/src/filesystem/implementation.js#L536) function definition

2.  only when a mode argument is passed: the `mode` is then associated with the new `Node`.
     if no argument, a mode value is still set the default file option of 0o644.

3.  a test has been created to ensure that if a mode is passed in as an argument for the [`fs.open`](https://github.com/filerjs/filer/blob/e3a1187ef94d5fc627768cee13457417e7ea0035/src/filesystem/implementation.js#L1610) function, the mode exists as it should. 

Would still be nice to... 
- test to ensure that mode's do not change when opening existing files, and passsing in a mode options.
- thinking of more ways to break this code would be helpful too.

### DIFF
--- a/src/filesystem/implementation.js
+++ b/src/filesystem/implementation.js
@@ -1924,7 +1924,6 @@ function isUint32(value) {
 // Validator for mode_t (the S_* constants). Valid numbers or octal strings
 // will be masked with 0o777 to be consistent with the behavior in POSIX APIs.
 function validateAndMaskMode(value, def, callback) {
-  console.log('Passed in mode: ' + value);
   if(typeof def === 'function') {
     callback = def;
     def = undefined;
@@ -1932,7 +1931,6 @@ function validateAndMaskMode(value, def, callback) {
 
   if (isUint32(value)) {
     let newMode = value & FULL_READ_WRITE_EXEC_PERMISSIONS;
-    console.log('Masked mode: ' + newMode);
     return value & FULL_READ_WRITE_EXEC_PERMISSIONS;
   }
 
@@ -1953,7 +1951,6 @@ function validateAndMaskMode(value, def, callback) {
       return false;
     }
     var parsed = parseInt(value, 8);
-    console.log('Parsed + Masekd mode: ' + parsed & FULL_READ_WRITE_EXEC_PERMISSIONS);
     return parsed & FULL_READ_WRITE_EXEC_PERMISSIONS;
   }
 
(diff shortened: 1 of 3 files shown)


humphd:
The mode is somewhat complicated by the fact that it includes two things: bits for the node's type (file, directory, link, ...); and also for permissions (e.g., 644, 777, etc).  So if you only want the permissions, you have to mask them out.

src/filesystem/implementation.js
```diff
@@ -533,7 +533,10 @@ function remove_directory(context, path, callback) {
   find_node(context, parentPath, read_parent_directory_data);
 }
 
-function open_file(context, path, flags, callback) {
+function open_file(context, path, flags, mode, callback) {
+  if (typeof mode == 'function'){
+    callback = mode;
```
humphd: You should also reset `mode` here, since it *was* a `function`, and now you're moving that over to `callback.  You should set it to `null`.

src/filesystem/implementation.js
```diff
@@ -645,6 +648,7 @@ function open_file(context, path, flags, callback) {
       }
       fileNode = result;
       fileNode.nlinks += 1;
+      if(mode){Node.setMode(mode, fileNode);}
```
humphd: Split this onto new lines please:

```js
if(mode) {
  Node.setMode(mode, fileNode);
}
```

steaward:
@humphd is there anything else I can do to make this PR a success?

modeswitch:
Sorry for the huge delay, I've started looking at this.

PR #477, opened 2018-09-24
Fix #470: Add test for fsPromises.symlink() when file soft link dose not exist

### LINKED ISSUE #470: Add test for fsPromises.symlink() when file soft link dose not exist
I found there is no promises version of the fs.symlink() test function.
And I would like to create one, make sure an error is received when the soft link of the file that does not exist.

@humphd

xwang345:
Added a new file to tet for fs.promises.symlink function

### DIFF
--- a/tests/spec/fs.symlink.spec.js
+++ b/tests/spec/fs.symlink.spec.js
@@ -43,4 +43,81 @@ describe('fs.symlink', function() {
       });
     });
   });
+
+  // it('should be return an error if the file destination path is bot exists', function(done){
+  //   var fsPromises = unil.fs().promises;
+
+  //   fsPromises.symlink('/tmp', '/', function(error){
+  //     expect(error).to.exist;
+  //     expect(error.code).to.equal('EEXIST');
+  //     done();
+  //   });
+  // });
+});
+
+describe('fs.promises.symlink', function(){
+  beforeEach(function(done){
+    util.setup(function(){
+      var fs = util.fs();
+      fs.promises.symlink('/','myfile')
+        .then(done())
+        .catch(function(error){
+          throw error;
+        });
+    });
+  });
+  afterEach(util.cleanup);
+
+  it('should be a function', function() {
+    var fs = util.fs();
+    expect(fs.promises.symlink).to.be.a('function');
+  });
+
+  // it('should be return an error if the file destination path is bot exists', function(done){
+  //   // var fs= unil.fs();
+
+  //   fs.promises.symlink('/tmp', '/', function(error){
+  //     expect(error).to.exist;
+  //     expect(error.code).to.equal('EEXIST');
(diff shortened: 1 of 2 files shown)


tests/spec/fs.symlink.spec.js
```diff
@@ -43,4 +43,81 @@ describe('fs.symlink', function() {
...
+    });
+  });
+  afterEach(util.cleanup);
+
+  it('should be a function', function() {
+    var fs = util.fs();
+    expect(fs.promises.symlink).to.be.a('function');
+  });
+
+  // it('should be return an error if the file destination path is bot exists', function(done){
```
ywpark1: I think you should uncomment this part since you want to test the case when the path does not exist.

ywpark1:
Under fsPromises.symlink section, there is no need to have tests for fs.symlink.

MarkKrutik:
The code has no problems whatsoever, however I have some styling tips that could make your code look even better.

You can call the promise API:
instead of var fs = util.fs().
That way you can distinguish fs.symlink() from fsPromises.symlink(). Otherwise very excellent work. Keep up the good work.

xwang345:
@MarkKrutik Thanks mention me to use util.fs().promises instead of util.fs(). I already implement into my test code.

xwang345:
@ywpark1 Yeah. You are right.

tests/spec/fs.symlink.spec.js
xwang345: Okay. I do this right now.

MarkKrutik:
No problem, happy to help!

tests/spec/fs.symlink.spec.js
```diff
@@ -44,3 +44,33 @@ describe('fs.symlink', function() {
     });
   });
 });
+
+describe('fs.promises.symlink', function(){
+  beforeEach(function(done){
+    util.setup(function(){
+      var fsPromises = util.fs().promises;
```
humphd: You can get rid of all this code, and just do `beforeEach(util.setup)`.  We don't need to do anything with `fs.symlink` before each test is run, only during the tests.

tests/spec/fs.symlink.spec.js
```diff
@@ -44,3 +44,33 @@ describe('fs.symlink', function() {
...
+  afterEach(util.cleanup);
+
+
+  it('should be a function', function() {
+    var fsPromises = util.fs().promises;
+    expect(fsPromises.symlink).to.be.a('function');
+  });
+
+
+  it('should return an error if the destination path already exists', () => {
```
humphd: This test description doesn't match what the test is doing.  Did you want to look for an error when the dest exists?  The code below is just symlinking `/` to `/tmp`, which will work with no errors.

PR #478, opened 2018-09-24
Fix #411: Tests if event is thrown when calling rename

### LINKED ISSUE #411: Add test for fs.watch.spec.js
There is no test looking for a `rename` event for fs.watch() in [fs.watch.spec.js](https://github.com/filerjs/filer/blob/master/tests/spec/fs.watch.spec.js)

A test case for this may be almost identical to [it('should get a change event when renaming a file', function(done) {](https://github.com/filerjs/filer/blob/fd3de6be2c1c19421d8c83e6791cd4b19afb84c7/tests/spec/fs.watch.spec.js#L43)

AHKol:
Could not make the `change` event to throw `rename` during a rename call.

Test will check if `change` event is thrown when a file is renamed.

### DIFF
--- a/tests/spec/fs.watch.spec.js
+++ b/tests/spec/fs.watch.spec.js
@@ -70,7 +70,7 @@ describe('fs.watch', function() {
 
     fs.writeFile('/myfile', 'data', function(error) {
       if(error) throw error;
-    })
+    });
 
     var watcher = fs.watch('/myfile', function(event, filename) {
       expect(event).to.equal('change');
--- a/tests/spec/fs.watch.spec.js
+++ b/tests/spec/fs.watch.spec.js
@@ -64,4 +64,23 @@ describe('fs.watch', function() {
       });
     });
   });
+
+  it('should get a change event when renaming a file', function(done) {
+    var fs = util.fs();
+
+    fs.writeFile('/myfile', 'data', function(error) {
+      if(error) throw error;
+    })
+
+    var watcher = fs.watch('/myfile', function(event, filename) {
+      expect(event).to.equal('change');
+      expect(filename).to.equal('/myfile');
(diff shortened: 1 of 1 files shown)


tests/spec/fs.watch.spec.js
```diff
@@ -64,4 +64,23 @@ describe('fs.watch', function() {
       });
     });
   });
+
+  it('should get a change event when renaming a file', function(done) {
+    var fs = util.fs();
+
+    fs.writeFile('/myfile', 'data', function(error) {
+      if(error) throw error;
+    })
```
0xazure: nit: Missing semi-colon.

Running `npm test` will catch a lot of these issues for you before you submit your PR.

0xazure:
@AHKol I'm having another look at this and I think I initially misunderstood the filer implementation. Did you see @humphd's [comment](https://github.com/filerjs/filer/issues/411#issuecomment-422946490) about the lack of `rename` events in filer? On NodeJS, `rename` events _are_ emitted when a file is renamed, so this test should actually be failing.

tests/spec/fs.watch.spec.js
```diff
@@ -64,4 +64,23 @@ describe('fs.watch', function() {
...
+
+  it('should get a change event when renaming a file', function(done) {
+    var fs = util.fs();
+
+    fs.writeFile('/myfile', 'data', function(error) {
+      if(error) throw error;
+    });
+
+    var watcher = fs.watch('/myfile', function(event, filename) {
+      expect(event).to.equal('change');
```
0xazure: This should be a `rename` event, which filer doesn't currently support.

AHKol:
@humphd Added the requested comment.

PR #481, opened 2018-09-24
Issue #436 - Implement basic fs.copyFile tests without flags

### LINKED ISSUE #436: Adding in support for fs.copyFile.
I noticed that Filer does not support [fs.copyFile](https://nodejs.org/api/fs.html#fs_fs_copyfile_src_dest_flags_callback). What's your opinion on me taking on writing out the code/doing a test for this? Difficult or good feature to work on? @humphd

Mordax:
Closes #436 kind of, implements non-flag testing for future fs.copyFile implementation.

Checks whether it's a function, checks if error is thrown if source that is being copied doesn't exist, checks if file was copied successfully.

### DIFF
--- a/tests/spec/fs.copyFile.spec.js
+++ b/tests/spec/fs.copyFile.spec.js
@@ -1,4 +1,3 @@
-var Filer = require('../../src');
 var util = require('../lib/test-utils.js');
 var expect = require('chai').expect;
 
@@ -37,7 +36,7 @@ describe('fs.copyFile', function(){
 
   it('should copy file successfully', function(done) {
     var fs = util.fs();
-    var src = "This is a src file.";
+    var src = 'This is a src file.';
 
     fs.copyFile('/srcfile', '/destfile', function(error) {
       if(error) throw error;
--- a/tests/spec/fs.copyFile.spec.js
+++ b/tests/spec/fs.copyFile.spec.js
@@ -1,0 +1,53 @@
+var Filer = require('../../src');
+var util = require('../lib/test-utils.js');
+var expect = require('chai').expect;
+
+describe('fs.copyFile', function(){
+  beforeEach(function(done){
+    util.setup(function() {
+      var fs = util.fs();
+      fs.writeFile('/srcfile', 'This is a src file.', function(error){
+        if(error) throw error;
+        fs.writeFile('/destfile', 'This is a dest file.', function(error){
+          if(error) throw error;
+          done();
+        });
+      });
+    });
+  });
+  afterEach(util.cleanup);
+
+  it('should be a function', function() {
+    var fs = util.fs();
+    expect(fs.copyFile).to.be.a('function');
+  });
+
+  it('should return an error if the src path does not exist', function(done){
+    var fs = util.fs();
+    var src = null;
+    var dest = 'dest.txt';
+
+    fs.copyFile(src, dest, function(error){
+      expect(error).to.exist;
+      expect(error.code).to.equal('ENOENT');
+      done();
+    });
+   
+  });
+
+  it('should copy file successfully', function(done) {
+    var fs = util.fs();
+    var src = "This is a src file.";
+
+    fs.copyFile('/srcfile', '/destfile', function(error) {
+      if(error) throw error;
+
+      fs.readFile('/destfile', function(error, data){
+        expect(error).not.to.exist;
+        expect(data).to.equal(src);
+        done();
+      });
+    });
+  });
+
+});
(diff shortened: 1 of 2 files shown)


Mordax:
Travis is supposed to fail due to no fs.copyFile implementation, just fyi.

tests/spec/fs.copyFile.spec.js
```diff
@@ -0,0 +1,52 @@
...
+
+    fs.copyFile(src, dest, function(error){
+      expect(error).to.exist;
+      expect(error.code).to.equal('ENOENT');
+      done();
+    });
+   
+  });
+
+  it('should copy file successfully', function(done) {
```
0xazure: This tests if `/destfile` has been overwritten with the contents of `/srcfile`, which is great! What about the case where `/destfile` does not already exist?

tests/spec/fs.copyFile.spec.js
```diff
@@ -0,0 +1,52 @@
...
+
+      fs.readFile('/destfile', function(error, data){
+        expect(error).not.to.exist;
+        expect(data).to.equal(src);
+        done();
+      });
+    });
+  });
+
+});
```
0xazure: nit: file should end with a blank line

tests/spec/fs.copyFile.spec.js
```diff
@@ -0,0 +1,52 @@
+var util = require('../lib/test-utils.js');
+var expect = require('chai').expect;
+
+describe('fs.copyFile', function(){
+  beforeEach(function(done){
+    util.setup(function() {
+      var fs = util.fs();
+      fs.writeFile('/srcfile', 'This is a src file.', function(error){
+        if(error) throw error;
+        fs.writeFile('/destfile', 'This is a dest file.', function(error){
```
0xazure: I would recommend isolating the creation of the `/destfile` to the test case that requires it to limit global state in the tests, especially if it's only used in one test. If it becomes necessary for multiple test cases, we can always move it back into the `beforeEach()` step.

tests/spec/fs.copyFile.spec.js
```diff
@@ -0,0 +1,52 @@
...
+
+  it('should be a function', function() {
+    var fs = util.fs();
+    expect(fs.copyFile).to.be.a('function');
+  });
+
+  it('should return an error if the src path does not exist', function(done){
+    var fs = util.fs();
+    var src = null;
+    var dest = 'dest.txt';
```
0xazure: nit: consistency - here you declare `dest` as a variable, but on [line 41](https://github.com/filerjs/filer/pull/481/files#diff-fef1ff9617a76f4b8c78fb20f986a75bR41) you use the filename directly in the function call. Since the destination filename is so short, I'd suggest putting it inline.

tests/spec/fs.copyFile.spec.js
```diff
@@ -0,0 +1,52 @@
+var util = require('../lib/test-utils.js');
+var expect = require('chai').expect;
+
+describe('fs.copyFile', function(){
```
humphd: Let's add `.skip()` to this, since we know we can't pass these tests yet.

https://mochajs.org/#inclusive-tests

tests/spec/fs.copyFile.spec.js
humphd: Agree, please move all these `writeFile` calls to the individual tests that need them.  It's best if tests have as few pre conditions as possible, and everything gets done in the test itself.

tests/spec/fs.copyFile.spec.js
humphd: Agree, inline this.  Also, paths in Filer must be absolute: `/dest.txt`

tests/spec/fs.copyFile.spec.js
```diff
@@ -0,0 +1,52 @@
...
+  afterEach(util.cleanup);
+
+  it('should be a function', function() {
+    var fs = util.fs();
+    expect(fs.copyFile).to.be.a('function');
+  });
+
+  it('should return an error if the src path does not exist', function(done){
+    var fs = util.fs();
+    var src = null;
```
humphd: This is prone to being incorrect changed in the future, since it looks like you just forgot to initialize your variable.  You need to document that you meant to do this.  A few options:

1. `var src; // intentionally left undefined`
1. `fs.copyFile(undefined /* intentionally undefined */, '/dest.txt', function(error) {`

tests/spec/fs.copyFile.spec.js
humphd: Add some extra info to line 37 to indicate `...when destination does not exist`

humphd:
I've rebased and fixed this, and landed in:

* https://github.com/filerjs/filer/commit/ef80713c3317bd0cf335cd736cf2a91534363650
* https://github.com/filerjs/filer/commit/62cbe11a9374c94f43f239232e95a41752a890fc
* https://github.com/filerjs/filer/commit/ba188169d663c44694e9304b2e1835d78c10825e

Issue #467, opened 2018-09-24
Add test for fs.chmod


rhayes2:
This test i don't think is written,

humphd:
@rhayes2 yeah, we need these.  The `fs.chmod` code is in, but there are zero tests.  Would be good to add it.

rhayes2:
having alot of issues setting things up on my non-main pc cause i had to go away, getting many git errors :O ugh

humphd:
If you want to hop on Slack and discuss, we can do that.  Or put some info in here?

humphd:
Here's one suggestion for how to test `fs.chmod`.

1. In this file https://github.com/filerjs/filer/blob/master/tests/spec/fs.chmod.spec.js
2. You want to add a new `it()` test block like the others
3. Add a test for the case that you pass an invalid `mode`, something like `1000`, and it should `expect` the `callback` to get an error or type `EINVAL`

PR #485, opened 2018-09-24
Issue 467 Adding Test for Chmod mode

rhayes2:
Added test for chmod  with an invalid mode comparison.  This will fix Issue #467

### DIFF
--- a/tests/spec/fs.chmod.spec.js
+++ b/tests/spec/fs.chmod.spec.js
@@ -44,7 +44,7 @@ describe('fs.chmod, fs.fchmod', function() {
 
     fs.mkdir('/file',function(err) {
       if(err) throw err;
-      
+        
       fs.chmod('file', 1000, function(err) {
         expect(err).to.exist;
         expect(err.code).to.equal('EINVAL');
--- a/tests/spec/fs.chmod.spec.js
+++ b/tests/spec/fs.chmod.spec.js
@@ -44,7 +44,7 @@ describe('fs.chmod, fs.fchmod', function() {
 
     fs.mkdir('/file',function(err) {
       if(err) throw err;
-        
+      
       fs.chmod('file', 1000, function(err) {
         expect(err).to.exist;
         expect(err.code).to.equal('EINVAL');
--- a/tests/spec/fs.chmod.spec.js
+++ b/tests/spec/fs.chmod.spec.js
@@ -39,19 +39,19 @@ describe('fs.chmod, fs.fchmod', function() {
     });
   });
 
-  it('Should be a valid mode value to set permissions', function(done){
+  it('should be a valid mode value to set permissions', function(done){
     var fs = util.fs();
-    fs.mkdir('/file',function(err, ed) {
+
+    fs.mkdir('/file',function(err) {
       if(err) throw err;
-      fs.chmod(ed, 1000, function(err) {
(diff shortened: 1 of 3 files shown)


tests/spec/fs.chmod.spec.js
```diff
@@ -39,6 +39,20 @@ describe('fs.chmod, fs.fchmod', function() {
     });
   });
 
+  it('Should be a valid mode value to set permissions', function(done){
```
humphd: Style nit: we tend to use a lower case letter on the test descriptions.  So use `should` vs. `Should`

tests/spec/fs.chmod.spec.js
```diff
@@ -39,6 +39,20 @@ describe('fs.chmod, fs.fchmod', function() {
...
+    var fs = util.fs();
+    fs.mkdir('/file',function(err, ed) {
```
humphd: The `ed` argument won't be returned by `mkdir`, it will only return an `err`.

tests/spec/fs.chmod.spec.js
```diff
@@ -39,6 +39,20 @@ describe('fs.chmod, fs.fchmod', function() {
...
+      if(err) throw err;
+      fs.chmod(ed, 1000, function(err) {
```
humphd: Instead of using `ed` here, use `'/file'` again

tests/spec/fs.chmod.spec.js
```diff
@@ -39,6 +39,20 @@ describe('fs.chmod, fs.fchmod', function() {
...
+        expect(err).to.exist;
+        expect(err.code).to.equal('EINVAL');
+        done();
+      });
+    });
+  });
+
+
```
humphd: You can remove these extra blank lines.

tests/spec/fs.chmod.spec.js
```diff
@@ -39,6 +39,20 @@ describe('fs.chmod, fs.fchmod', function() {
     });
   });
 
+  it('should be a valid mode value to set permissions', function(done){
+    var fs = util.fs();
+
+    fs.mkdir('/file', function(err) {
```
humphd: This `mkdir` call isn't necessary.  You can just use `/` as your directory (always exists).

tests/spec/fs.chmod.spec.js
```diff
@@ -39,6 +39,20 @@ describe('fs.chmod, fs.fchmod', function() {
...
+      if(err) throw err;
+      
+      fs.chmod('file', 1000, function(err) {
```
humphd: This is valid in node.js (i.e., you can use a mode of `1000`).  Same for `"1000"`, which node will turn into an integer.  If you want to make this test fail, you could use `{}`:

```js
fs.chmod('file', {}, function(err) {
  expect(err).to.exist;
  expect(err.code).to.equal('EINVAL');
  done();
});
```

node will give you a `TypeError [ERR_INVALID_ARG_TYPE]: The "mode" argument must be of type number. Received type object` which we should probably match.

humphd:
I finished this in https://github.com/filerjs/filer/commit/acef2c93b0008610b97c58926ba26909cb680a2a, rebased, and added two fixes for failing test cases that this found.  Nice work.

PR #486, opened 2018-09-24
Fix #450 Add test for fs.writeFile-readFile.spec to write a file in non-existi…

### LINKED ISSUE #450: Add test for fs.writeFile-readFile.spec to write a file in non-existing path
I wang to add a test for fs.writeFile-readFile.spec to check writing a file in non-existing path. I would like to work on this issue. Could you please assign me to this test?

giantpanpan:
I add a test for fs.writeFile-readFile.spec to write a file in non-existing path.

### DIFF
--- a/tests/spec/fs.writeFile-readFile.spec.js
+++ b/tests/spec/fs.writeFile-readFile.spec.js
@@ -22,6 +22,19 @@ describe('fs.writeFile, fs.readFile', function() {
     });
   });
 
+  
+  it('should error when path is wrong to writeFile',function(done){
+    var fs = util.fs();
+
+    fs.writeFile('/tmp/myfile', '','utf8', function(error, result) {
+      expect(error).to.exist;
+      expect(error.code).to.equal('ENOENT');
+      expect(result).not.to.exist;
+      done();
+    });
+
+  });
+
   it('should write, read a utf8 file without specifying utf8 in writeFile', function(done) {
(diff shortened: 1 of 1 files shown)


tests/spec/fs.writeFile-readFile.spec.js
```diff
@@ -22,6 +22,19 @@ describe('fs.writeFile, fs.readFile', function() {
     });
   });
 
+  
+  it('should error when path is wrong to writeFile',function(done){
```
YuechengWu: Just to nitpick here, you can use ()=>{....} to declare a function instead of function().

tests/spec/fs.writeFile-readFile.spec.js
```diff
@@ -22,6 +22,19 @@ describe('fs.writeFile, fs.readFile', function() {
...
+    var fs = util.fs();
+
+    fs.writeFile('/tmp/myfile', '','utf8', function(error, result) {
+      expect(error).to.exist;
+      expect(error.code).to.equal('ENOENT');
+      expect(result).not.to.exist;
+      done();
+    });
+
```
YuechengWu: We can also try use Promises to write the test, for instance:
it('(promise) should return an error when path is wrong to writeFile', ()=>{
return fsPromises.writeFile (......)
.catch(error => { expect...... }); // and you don't need done() at the end.
});

PR #492, opened 2018-09-24
add test for fs.appendFile to check when symbolic link does not exist

shreenaathia:
add test for fs.appendFile to check when symbolic link does not exist


tests/spec/fs.appendFile.spec.js
```diff
@@ -110,6 +110,19 @@ describe('fs.appendFile', function() {
...
   });
 
+  it('should error when the symbolic link does not exist', function(done) {
+    var fs = util.fs();
+    var contents = 'This is a file.';
+     
+    fs.appendFile('/newfile', contents, { encoding: 'utf8' }, function(error) {  
+    expect(error).to.exist;
+    expect(error.code).to.equal('ENOENT');
+    expect(result).not.to.exist;
```
VinceBackpack: `result` is being used in here but it has not been declared.

DinahLin: In my opinion, ENOENT means the file or directory is not exist, so you do not need to write  120 line.

shreenaathia: Thank you! I am changing that, 
Must be more careful when I am using other test examples.

shreenaathia: Thank you for your input, I did not need the line at 120, and have changed actually ENOENT to EINVAL to pass this test.

tests/spec/fs.mkdir.spec.js
```diff
@@ -46,3 +46,5 @@ describe('fs.mkdir', function() {
     });
   });
 });
+
```
humphd: You can undo these changes to an unrelated file like this: `git checkout master tests/spec/fs.mkdir.spec.js`

tests/spec/fs.appendFile.spec.js
```diff
@@ -110,11 +110,22 @@ describe('fs.appendFile', function() {
     });
   });
 
-  it('should work when file does not exist, and create the file', function(done) {
+  it('should error when the symbolic link does not exist', function(done) {
+    var fs = util.fs();
+     
+    fs.symlink('/myfile', ' ', function(error) {
```
humphd: We want to avoid putting unrelated tests in a file with other functions.  Here you've got a test for `symlink` in a file for tests of `appendFile`.  Did you want to try and rewrite this in terms of something to do with `appendFile`?  If not, do you want to move this test to the `symlink` test file?

tests/spec/fs.appendFile.spec.js
```diff
@@ -110,11 +110,22 @@ describe('fs.appendFile', function() {
...
+    });
+  });
+
+
+  it('should work when file does not exist, and create the file',function(done) {
     var fs = util.fs();
     var contents = 'This is a file.';
 
-    fs.appendFile('/newfile', contents, { encoding: 'utf8' }, function(error) {
+    fs.appendFile('/newfile', contents, { encoding: 'utf8' },function(error) {
```
humphd: Be careful with changes like this one, which only change whitespace.  Projects tend to not want them, since they end up hiding history in git (i.e., it looks like you altered this file, but really, the code was introduced before you touched this line).

Issue #463, opened 2018-09-23
add tests for fs.appendFile for valid file system flags


MuchtarSalimov:
When calling [fs.appendFile(path, data[, options], callback)](https://nodejs.org/docs/latest/api/fs.html#fs_fs_appendfile_path_data_options_callback), within the options object we may include a file system flag (defaults to 'a' so that the file is opened for appending).

I would like to:

- add a test for explicitly setting the mode to 'a' when using fs.appendFile.

However it is not obvious to me why the user might have the option of choosing a flag other than 'a'. Is there any value in leaving it as an option? It seems to me that for this method, any selection would either cause it to behave as if it were opened using 'a', or else cause some kind of error.

humphd:
See https://nodejs.org/api/fs.html#fs_file_system_flags and https://github.com/nodejs/node/blob/eef072fa083f05f84fa6ca1908472eb228095a38/lib/fs.js#L1227-L1238 for what node does.  Filer needs to support the same options as node, even if we end up ignoring the.  I agree that if you pass certain flags it should fail, but that's a reasonable test to write.  My advice: try passing different flags to node's version, see what it does, write tests for Filer that expect the same behaviour.

PR #494, opened 2018-09-24
Add test cases for fs.appendFile with explicit encoding and flag options

MuchtarSalimov:
Originally intending to add a test case for [fs.appendFile](https://nodejs.org/api/fs.html#fs_fs_appendfile_path_data_options_callback) to verify that selecting a flag manually does work as intended (#463), it seems that that test does in fact fail in some cases, but not in others. All calls to fs.appendFile here are tested with what should be their default values. I have instead added two tests instead of one, in order to illustrate where fs.appendFile seems to fail:

fs.appendFile(path, data[, options], callback)

New test 1 (passes) :   options ={ encoding: 'utf8', flag: 'a' }

New test 2 (fails): options = { flag: 'a'}

Similar existing test (passes): options = { encoding: 'utf8' }

### DIFF
--- a/tests/spec/fs.appendFile.spec.js
+++ b/tests/spec/fs.appendFile.spec.js
@@ -50,6 +50,38 @@ describe('fs.appendFile', function() {
     });
   });
 
+  it('should append without error when explcitly entering encoding and flag options (default values)' , function(done) {
+    var fs = util.fs();
+    var contents = 'This is a file.';
+    var more = ' Appended.';
+
+    fs.appendFile('/myfile', more , {encoding: 'utf8', flag: 'a'}, function(error) {
+      if(error) throw error;
+
+      fs.readFile('/myfile', { encoding: 'utf8' }, function(error, data) {
+        expect(error).not.to.exist;
+        expect(data).to.equal(contents + more);
+        done();
+      });
+    });
+  });
+
+  it('should append without error when specfifying flag option (default value)' , function(done) {
+    var fs = util.fs();
+    var contents = 'This is a file.';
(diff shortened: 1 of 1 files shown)


tests/spec/fs.appendFile.spec.js
```diff
@@ -50,6 +50,38 @@ describe('fs.appendFile', function() {
     });
   });
 
+  it('should append without error when explcitly entering encoding and flag options (default values)' , function(done) {
+    var fs = util.fs();
+    var contents = 'This is a file.';
+    var more = ' Appended.';
+
+    fs.appendFile('/myfile', more , {encoding: 'utf8', flag: 'a'}, function(error) {
```
YuechengWu: Hi just to nitpick here, did you try using promises instead of done()?
For instance, declare var fsPromises = util.fs().promises;
then return fsPromises.appendFile(......) .catch (error => {......})
also, fsPromises.readFile(......) .catch(error => {expect(error)......}), and you don't need done() at the end.

MuchtarSalimov: Wouldn't that be testing a different function? I'm not sure I understand why this should be done.

humphd: We could do the test in both ways, it's true.  Using a `done` callback is fine if you want to use callbacks, and if you're doing Promises, not needed.  In this cas @YuechengWu is using callbacks, so this should be OK>

rhayes2: That's an interesting bug you found in class!

PR #495, opened 2018-09-24
Create issue-473

ksandhu16:
Adding test for fs.writeFile() when the string is externalized.
Hello! I have attempted in adding an fs.writeFile() test. Any feedback would be appreciated.


struong9:
Although it's not code related, it would be good to include what this Push Request is trying to address in the title.

PR #496, opened 2018-09-24
fix #418 test for fsPromises.truncate(path[, len])

### LINKED ISSUE #418: Add test for fsPromises.truncate(path[, len]) to test when length is negative
There are no test available for [fsPromises.truncate(path[, len])](https://nodejs.org/api/fs.html#fs_fspromises_truncate_path_len) .

I would like to work on the test method to check if the [length of the file is negative](https://github.com/filerjs/filer/blob/master/tests/spec/fs.truncate.spec.js#L13) .

Please let me know if it is okay to work.

deepanjali19:
This fixes the issue #418.

There is now a test available for fsPromises.truncate(path[, len]) .

It checks if the length of the file is negative, and sends an error message.

### DIFF
--- a/tests/spec/fs.truncate.spec.js
+++ b/tests/spec/fs.truncate.spec.js
@@ -192,3 +192,27 @@ describe('fs.truncate', function() {
     });
   });
 });
+
+describe('fsPromises.truncate', function () {
+  beforeEach(util.setup);
+  afterEach(util.cleanup);
+
+  it('should error when length is negative', () => {
+
+    var fsPromises = util.fs().promises;
+    var contents = 'This is a file.';
+
+    fsPromises.writeFile('/myfile', contents)
+      .then(
+        error => {
+          if (error) throw error;
+        });
+
+    fsPromises.truncate('/myfile', -1)
+      .then(
+        error => {
+          expect(error).to.exist;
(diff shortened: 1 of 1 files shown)


tests/spec/fs.truncate.spec.js
```diff
@@ -192,3 +192,21 @@ describe('fs.truncate', function() {
...
+describe('fsPromises.truncate', function () {
+  beforeEach(util.setup);
+  afterEach(util.cleanup);
+
+  it('should error when length is negative', () => {
+
+    var fsPromises = util.fs().promises;
+    var contents = 'This is a file.';
+
+    return fsPromises.writeFile('/myfile', contents)
```
biskit1: Does this need to be a return @deepanjali19?

humphd: It does, so that the containing test gets a Promise it can use to know when the test is complete.  With callback based tests you don't need to; but with Promise based ones you do.

deepanjali19: Thank you @humphd for explaining it better than I could have.

humphd:
@deepanjali19 this looks good, we just need to resolve this merge conflict.  Let's do it in class tomorrow together.

deepanjali19:
Sure @humphd !

PR #498, opened 2018-09-24
Enforce newline at end of files

0xazure:
Trailing newlines are a common idiom in programming, and enabling this rule will save reviewers nitpicking or overlooking missing trailing newlines.

ESLint also supports the `--fix` option to automate fixing some of the problems reported by this rule.

There are no current files in the codebase that require fixing to add trailing newlines as verified by `npm run lint`.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -12,6 +12,10 @@
         "sourceType": "module"
     },
     "rules": {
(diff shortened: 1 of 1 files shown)


humphd:
I'll let @modeswitch review this one.

PR #501, opened 2018-09-24
491

yevseytsev:
mkdir - restricted symbols ('?' only)

### DIFF
--- a/tests/spec/fs.mkdir.spec.js
+++ b/tests/spec/fs.mkdir.spec.js
@@ -45,4 +45,16 @@ describe('fs.mkdir', function() {
       });
     });
   });
+
+  //not sure about error code
+  it('should not create dir with restricted symbols', function(done) {
+    var fs = util.fs();
+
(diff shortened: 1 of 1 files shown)


tests/spec/fs.mkdir.spec.js
```diff
@@ -45,4 +45,16 @@ describe('fs.mkdir', function() {
...
+  it('should not create dir with restricted symbols', function(done) {
+    var fs = util.fs();
+
+    fs.mkdir('/?', function(error) {
+      expect(error).to.exist;
+      expect(error.code).to.equal('DIRECTORY');
+      done();
+    });
+  });
+
```
YuechengWu: Hi, just to nitpick here,
Maybe we can try throwing promises for example:
it('(promise) should not create dir with restricted symbols', () => {
  return fsPromises.mkdir('/?')
    .catch(error => {
      expect(error).to.exist;
    });
});

PR #504, opened 2018-09-24
Issue 456

DinahLin:
If it is the success, the filename is undefined, when we write a file from original to a destination.
I change code at fs.write.spec.js


tests/spec/fs.write.spec.js
```diff
@@ -5,6 +5,18 @@ var expect = require('chai').expect;
 describe('fs.write', function() {
   beforeEach(util.setup);
   afterEach(util.cleanup);
+  
+  // the file is underfined, will create problem
```
AHKol: Typo nit: 'unde**r**fined' might be intended to spell 'undefined'.

DinahLin: Thank you for telling me, I will fix the problem late.

tests/spec/fs.write.spec.js
```diff
@@ -5,6 +5,18 @@ var expect = require('chai').expect;
 describe('fs.write', function() {
   beforeEach(util.setup);
   afterEach(util.cleanup);
+  
+  // if the file is undefined, it will create problems
+  // 
+  it('should be a error',function(done){
```
humphd: Please move your comment above into the test message like this:

```js
it('should error if the file name is undefined', function(done) {
```

humphd:
I've fixed this up, rebased, and landed in https://github.com/filerjs/filer/commit/4a5d8c1222cb454040b28397896bbf667ba6b2fe with some extra fixes in https://github.com/filerjs/filer/commit/158f6241b6fb437c1cae6522085d0db7c051f49b.

PR #505, opened 2018-09-24
Fixes #499: Add a test for truncate when length is NaN

### LINKED ISSUE #499: Add a test for truncate when length is NaN
Let's consider adding a test for when `fs.truncate`'s `length` parameter is `NaN` (ie. Not a number like `stringInputForTruncateExplodes`).

The tests live within [`filer/tests/spec/fs.truncate.spec.js`](https://github.com/filerjs/filer/blob/master/tests/spec/fs.truncate.spec.js) and could possibly slide right under the test [`when length is negative`](https://github.com/filerjs/filer/blob/e3a1187ef94d5fc627768cee13457417e7ea0035/tests/spec/fs.truncate.spec.js#L13)!

[Node.js docs](https://nodejs.org/api/fs.html#fs_fspromises_truncate_path_len)

SeanPrashad:
Fixes #499 - Add a test for truncate when length is NaN

This change to the  functionality allows fs.truncate() to accept a value for
length that is NaN (Not a Number) and invoke an appropriate error.

MDN Documentation for `Number.IsInteger()` found [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger).

### DIFF
--- a/tests/spec/fs.truncate.spec.js
+++ b/tests/spec/fs.truncate.spec.js
@@ -25,6 +25,21 @@ describe('fs.truncate', function() {
     });
   });
 
+  it('should error when length is NaN', function(done) {
+    var fs = util.fs();
+    var contents = 'This is a file.';
+
+    fs.writeFile('/myfile', contents, function(error) {
+      if(error) throw error;
+
+      fs.truncate('/myfile', 'NaN', function(error) {
+        expect(error).to.exist;
+        expect(error.code).to.equal('EINVAL');
+        done();
+      });
+    });
+  });
+
   it('should error when path is not a file', function(done) {
     var fs = util.fs();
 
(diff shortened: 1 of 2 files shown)


tests/spec/fs.truncate.spec.js
```diff
@@ -25,6 +25,21 @@ describe('fs.truncate', function() {
     });
   });
 
+  it('should error when length is NaN', function(done) {
+    var fs = util.fs();
```
Janice0218: It maybe naming variables like fsPromises instead of fs() that is make more clear.

SeanPrashad:
**Note**: Closing this PR as I've reopened a new one for demonstration purposes during my blog post.

tests/spec/fs.truncate.spec.js
SeanPrashad: Nice catch @Janice0218 - I chose not to use `fsPromises` since I'm not using the asynchronous version of it. I should probably reopen a new PR to focus on `fsPromises.truncate` instead 🤔

PR #507, opened 2018-09-24
Fix #500: add test for fs.open to check flags

### LINKED ISSUE #500: Add test for fs.open when flag in invalid
I noticed that there is no test to check that an error is thrown when the flag in invalid for fs.open. 
A test can be added for this case.

I would like to be assigned to this issue.

Mera-Gangapersaud:
Added test for fs.open to check that flags are valid per issue #500

### DIFF
--- a/tests/spec/fs.open.spec.js
+++ b/tests/spec/fs.open.spec.js
@@ -136,4 +136,13 @@ describe('fs.open', function() {
       });
     });
   });
+
+  it('should return an error when flag in invalid', function(done) { 
+    var fs = util.fs();
+
+    expect(fs.open('/myfile', 'abcd', function(){
+      done();
+    })).to.throw('flags is not valid');
+    
+  });
 });


SeanPrashad:
👏🏽 This 👏🏽 is 👏🏽 some 👏🏽 good 👏🏽 stuff 👏🏽 @Mera-Gangapersaud!

It looks like we have one nit (nit-pick, ie. something small) to fix up according to TravisCI:

https://travis-ci.org/filerjs/filer/builds/432704203#L2181

When you have a chance, feel free to fix the linting issue, then you can either add a new commit with a message of something like `Correct TravisCI linting error`, or (what I would do) **append** your fix to your previous commit. This will essentially tack that one linting fix onto your existing commit (https://github.com/filerjs/filer/pull/507/commits/79afd62364ebb7f49ac8b3f42dccd087cd07f4cc) and it'll cause TravisCI to rerun!

To do the second method, you'll need to:

(1) Correct the linting issue in your editor of choice
(2) Enter `git status` in your terminal/command prompt and verify that single file was the only thing that changed
(3) Enter `git add fileThatIModifiedToFixTravis` to add the file to staging
(4) Enter `git commit --amend` to **add that fix to the last commit**

*And voila!* 🎉 

Give me a shout if you need any help with this!

SeanPrashad:
Also, we'll need to include the original issue number (#500) in either the description, or commit message so that it automatically closes the respective issue!

SeanPrashad:
I see the linting issue is fixed - nice work!!

We now have a new problem (_that isn't soo bad.._). Since we amended that linting fix to the previous commit, you did a good thing and updated the commit message. However, this over wrote the old commit message, which was more descriptive and what we were looking for.

Can you run `git commit --amend` and update it to be something like the original message you had? Maybe something like `Fixes #500: Add a test for fs.open() when flag is invalid`

tests/spec/fs.open.spec.js
```diff
@@ -136,4 +136,13 @@ describe('fs.open', function() {
       });
     });
   });
+
+  it('should return an error when flag in invalid', function(done) {
```
humphd: Let's change `return` in this to `throw`

tests/spec/fs.open.spec.js
```diff
@@ -136,4 +136,13 @@ describe('fs.open', function() {
       });
     });
   });
+
+  it('should return an error when flag in invalid', function(done) { 
+    var fs = util.fs();
+
+    expect(fs.open('/myfile', 'abcd', function(){
+      done();
```
humphd: I don't think we need `done` at all in this test, since nothing async is happening (i.e., it throws right away

tests/spec/fs.open.spec.js
```diff
@@ -136,4 +136,13 @@ describe('fs.open', function() {
       });
     });
   });
+
+  it('should return an error when flag in invalid', function(done) { 
+    var fs = util.fs();
+
+    expect(fs.open('/myfile', 'abcd', function(){
```
humphd: Usually you can't pass a function that will throw like this, you have to wrap it in a function that the test runner can call:

```js
expect(function() {
  fs.open('/myfile', 'abcd', function(){});
}).to.throw('flags is not valid');
```

It's sort of like having a fuse on a bomb: before you set a bomb off, you need to be far away.  By wrapping the function in an outer function, you can safely call the outer function, knowing that it will likely throw.

humphd:
I've made the changes above, rebased, and in doing so, discovered that we have a bug that this test revealed.  Nice work @Mera-Gangapersaud.  I added a fix in https://github.com/filerjs/filer/commit/97d2d1bfe3a5e729f78684e6e688beb870c36bcc.  Your code is landed now.

Issue #483, opened 2018-09-24
Add tests for fsPromises.access


mskuybeda:
currently there are no tests involving the promises version of the fs.access() function.

I would like to work on it.
Thanks,

Nick S

mskuybeda:
I would like to change my function to fsPromises.mknod since fs.access was not implemented.

humphd:
Sounds good.

PR #509, opened 2018-09-24
Fixed Issue #483 for a test done for fs.mknod promise

mskuybeda:
used the existing test for fs and adjusted it to promise based version. 
it creates a directory to test a node at path

### DIFF
--- a/tests/spec/fs.mknod.spec.js
+++ b/tests/spec/fs.mknod.spec.js
@@ -1,64 +1,64 @@
 var util = require('../lib/test-utils.js');
 var expect = require('chai').expect;
 
-describe('fs.mknod', function () {
+describe('fs.mknod', function(){
   beforeEach(util.setup);
   afterEach(util.cleanup);
 
-  it('should be a function', function (done) {
+  it('should be a function', function(done){
     var fs = util.fs();
     expect(fs.mknod).to.be.a('function');
     done();
   });
 
-  it('should return an error if part of the parent path does not exist', function (done) {
+  it('should return an error if part of the parent path does not exist', function(done){
     var fs = util.fs();
 
-    fs.mknod('/dir/mydir', 'DIRECTORY', function (error) {
+    fs.mknod('/dir/mydir', 'DIRECTORY', function(error){
       expect(error.code).to.equal('ENOENT');
       done();
     });
   });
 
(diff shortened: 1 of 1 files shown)


tests/spec/fs.mknod.spec.js
```diff
@@ -1,82 +1,94 @@
...
+      if (error) throw error;
+
+      fs.stat('/file', function (error, result) {
         expect(error).not.to.exist;
         expect(result.type).to.equal('FILE');
         done();
       });
     });
   });
+  describe.only('Promise test mknod', function () {
```
SeanPrashad: I think it's OK to keep your newly added test inside the existing `describe` block - other students have been putting them there!

tests/spec/fs.mknod.spec.js
```diff
@@ -1,82 +1,94 @@
 var util = require('../lib/test-utils.js');
 var expect = require('chai').expect;
 
-describe('fs.mknod', function() {
+describe('fs.mknod', function () {
```
SeanPrashad: Let's consider cleaning up the white spaces in a separate PR! We should aim to keep PR's confined to one specific thing - in this case, it's simply adding a new test as stated in #483. Another issue should be opened with a title of `Remove whitespace in test files` if you'd like!

With that being said, can you please revert the white space changes in a new commit? Please let me know if you need any help with this! 😁

tests/spec/fs.mknod.spec.js
```diff
@@ -1,82 +1,94 @@
...
         expect(result.type).to.equal('FILE');
         done();
       });
     });
   });
+  describe.only('Promise test mknod', function () {
+
+    it('should return an error if part of the parent path does not exist', function () {
+      var fs = util.fs().promises;
+
```
SeanPrashad: We can remove this blank line ✌️

mskuybeda: solved

SeanPrashad:
@mskuybeda I definitely agree - this calls for [Prettier](https://prettier.io/)

humphd:
@mskuybeda I'm going to land this, but it's not good that you modified all those other lines for style-only changes, and then didn't reset them.  In the future, be very careful about this, and don't let this type of thing get into your PRs.

PR #511, opened 2018-09-24
Fix issus503: add promises version of close method

qwang135:


tests/spec/fs.close.spec.js
```diff
@@ -31,3 +31,13 @@ describe('fs.close', function() {
...
+
+describe('fsPromises.close', function() {
+  beforeEach(util.setup);
+  afterEach(util.cleanup);
+
+  it('should be a function', function() {
+    var fsPromises = util.fs().promises;
+    expect(typeof fsPromises).to.equal('function');
+  });
+});
```
thanhcng: Here is a working test case for[ should release the file handler](https://github.com/filerjs/filer/blob/e3a1187ef94d5fc627768cee13457417e7ea0035/tests/spec/fs.close.spec.js#L14) test case. Here we are not working with file descriptor when writing `fsPromises` code, but instead we're dealing with [FileHandle ](https://nodejs.org/api/fs.html#fs_class_filehandle)object, which is a wrapper object for file descriptor. And for our test, were're not using `fsPromises` to close `FileHandle`, but using `FileHandle`'s `close()` method and `fd` property instead. 
```
it('should release the file handler', function() {
    var buffer = new Filer.Buffer(0);
    var filehandle;

    return fsPromises.open('/myfile', 'w')
      .then((fh) => {
        filehandle = fh;
        return filehandle.close();
      })
      .then(() => {
        filehandle.read(buffer, 0, buffer.length, undefined);
      })
      .catch(err => {
        expect(err).to.exist;
        expect(filehandle.fd).not.to.exist;
      });
});
```

humphd:
The `fs.close` method got removed from the list of promise versions since you wrote this.  Sorry for the confusion (it's my fault).  But this isn't needed now.

PR #516, opened 2018-09-25
Fixed issue#480 using promises to test fs.truncate when path does not exist

### LINKED ISSUE #480: Add a test for fs.truncate() when path doesn't exist
Hi,

I have noticed that there is currently no test for fs.truncate() when the path does not exist.
If that's okay and there is indeed no current test for this, I would like to work on this issue, please.

Please let me know.
Thank you

ooHAoo:
Fixes #480  - Add a test for truncate when path does not exist using promises.

I attempted to add a test for when fs.truncate is being passed a path/file that does not exist.
This was all done using promises.

### DIFF
--- a/tests/spec/fs.truncate.spec.js
+++ b/tests/spec/fs.truncate.spec.js
@@ -192,3 +192,22 @@ describe('fs.truncate', function() {
     });
   });
 });
+
+
+describe('fsPromises.truncate', function () {
+  beforeEach(util.setup);
+  afterEach(util.cleanup);
+  it('should error when path does not exist (with promises)', (done) => {
+    var fsPromises = util.fs().promises;
+
+    fsPromises.truncate('/NonExistingPath', 0)
+      .then(result => expect(result).not.to.exist)
+      .catch(error => {
+        expect(error).to.exist;
+        expect(error.code).to.equal('ENOENT');
+      })
+      .then(() => done());
+  });
+});
+
+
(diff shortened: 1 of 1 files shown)


tests/spec/fs.truncate.spec.js
```diff
@@ -192,3 +192,22 @@ describe('fs.truncate', function() {
     });
   });
 });
+
+
+describe('fsPromises.truncate', function () {
+  beforeEach(util.setup);
+  afterEach(util.cleanup);
+  it('should error when path does not exist (with promises)', (done) => {
```
humphd: You can get rid of `done` in this callback

tests/spec/fs.truncate.spec.js
```diff
@@ -192,3 +192,22 @@ describe('fs.truncate', function() {
...
+    var fsPromises = util.fs().promises;
+
+    fsPromises.truncate('/NonExistingPath', 0)
```
humphd: Add a `return` before the `fsPromises...` here, so you return the Promise for this test, to signal it's done.

tests/spec/fs.truncate.spec.js
```diff
@@ -192,3 +192,22 @@ describe('fs.truncate', function() {
...
+      .then(result => expect(result).not.to.exist)
```
humphd: I think you can delete this line.

biskit1: Should this be a `.catch(...)` error check instead?

PR #519, opened 2018-09-25
Fix issue #475 - Add test for fs.exits() when resulting path of a symbolic link is not found

### LINKED ISSUE #475: Adding test for fs.exists() when it follows symbolic link and can't find directory
I have noticed the test **fs.exists()** doesn't account for the case when the the path is a symbolic link and there is no resulting directory.

I would like to add the test for that. What do you think?

Alternatively I could also create a test for **fs.rmdir()** to test when the target is a (hard) link, which the test should fail.

@humphd

iblacker:
Fixes #475 - Add test for fs.exists() when it follows symbolic link and can't find directory.

It tests by creating a symbolic link for a path and then deleting the path. The test should return false.

Passed all the tests

### DIFF
--- a/tests/spec/fs.exists.spec.js
+++ b/tests/spec/fs.exists.spec.js
@@ -56,4 +56,28 @@ describe('fs.exists', function() {
       });
     });
   });
+  it('should follow symbolic links and return false if for the resulting path does not exist', function(done) {
+    var fs = util.fs();
+
+    fs.open('/myfile', 'w', function(error, fd) {
+      if(error) throw error;
+
+      fs.close(fd, function(error) {
+        if(error) throw error;
+
+        fs.symlink('/myfile', '/myfilelink', function(error) {
+          if(error) throw error;
+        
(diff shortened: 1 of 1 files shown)


tests/spec/fs.exists.spec.js
```diff
@@ -56,4 +56,28 @@ describe('fs.exists', function() {
       });
     });
   });
+  it('should follow symbolic links and return false if for the resulting path does not exist', function(done) {
```
humphd: Add a blank line above this, so it visually separates this test from the others.

humphd: Great question.  To update a PR, you simply `add` and `commit` again on your same `branch`, then `push` your `branch` again to your `origin`.  GitHub will add that commit (or commits) to this PR.

Let me know if you have any trouble or need more help.

humphd:
@iblacker you did that perfectly, nice work.

PR #525, opened 2018-09-26
Issue497 Add test for fs.chmod when the path is invalid

yuzhouChen:
Added a function to test the [fs.chomd](https://github.com/filerjs/filer/blob/master/tests/spec/fs.chmod.spec.js), when people passed the invalid path as the argument, they will get the error code 'ENOENT'. All tests are passed.

### DIFF
--- a/tests/spec/fs.chmod.spec.js
+++ b/tests/spec/fs.chmod.spec.js
@@ -39,6 +39,17 @@ describe('fs.chmod, fs.fchmod', function() {
     });
   });
 
+  it('should error whenp path is invalid', function(done){
+    var fs = util.fs();
+
+      fs.chmod("/invalid_file", 0o444, function(err) {
+        expect(err).to.exist;
+        expect(err.code).to.equal('EINVAL');
+        done();
+      });
+    });
+  });
+
   it('should allow for updating mode of a given file', function(done) {
     var fs = util.fs();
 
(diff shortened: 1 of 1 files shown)


tests/spec/fs.chmod.spec.js
```diff
@@ -39,6 +39,16 @@ describe('fs.chmod, fs.fchmod', function() {
     });
   });
 
+  it('should be an error when the path is invalid', function(done){
```
MuchtarSalimov: I think the wording here should be more about the method _doing_ something rather than _being_ something. Perhaps something like "it should throw an error" would be better?

Having taken a look at existing tests, there is a lot of variation on how people have chosen to title the tests. I've seen both "it should fail" and "it should return an error" in various places.

tests/spec/fs.chmod.spec.js
```diff
@@ -39,6 +39,16 @@ describe('fs.chmod, fs.fchmod', function() {
...
+    var fs = util.fs();
+    fs.chmod('/invalid_path', 0o444, function(err){
+      expect(err).to.exist;
+      expect(err.code).to.equal('ENOENT');
+      done();
+    });
+
```
MuchtarSalimov: Nitpick: Extra line of whitespace here doesn't match other tests.

Issue #529, opened 2018-09-26
Add test for fsPromises.lstat()


Janice0218:
I found no test for fs.promises.lstat() for these two cases: "should be a function" and "should return an error if path does not exist", I would like you to assign me this issue.

jrkong:
@qwang135 @Janice0218 you two filed the same issue #514. Maybe the two of you can work together?

PR #530, opened 2018-09-26
Fix issue 529

Janice0218:
This pull request [fixes ](https://github.com/Janice0218/filer/blob/85b6978393a905c48d655f945ded47b8495aa0e3/tests/spec/fs.lstat.spec.js#L51) issue #529


tests/spec/fs.lstat.spec.js
```diff
@@ -47,3 +47,29 @@ describe('fs.lstat', function() {
     });
   });
 });
+
+describe('fs.promises.lstat', function() {
```
xwang345: Just nitpick there, you can wirte a function like () => instead of function(). The arrow function expression has a shorter syntax than a function expression.

SeanPrashad: Good suggestion @xwang345 - in a case like this, I'd vouch to not incorporate arrow functions since other tests don't use that syntax. Consistency across code goes a long way!

tests/spec/fs.lstat.spec.js
```diff
@@ -47,3 +47,29 @@ describe('fs.lstat', function() {
...
+  it('should be a function', function() {
+    var fsPromises = util.fs().promises;
+    expect(typeof fsPromises.lstat).to.equal('function');
+  });
+
+  it('should return an error if path does not exist', function() {
+    var fsPromises = util.fs().promises;
+    var fsStats;
+    fsPromises.lstat('/tmp')
+      .then( (stat) => {
```
SeanPrashad: One too many spaces 😁 - this should be `.then((stat) => { `, or in this case you don't need the parenthesis (`(` and `)`) because you only have one parameter!

tests/spec/fs.lstat.spec.js
```diff
@@ -47,3 +47,29 @@ describe('fs.lstat', function() {
...
+
+  it('should be a function', function() {
+    var fsPromises = util.fs().promises;
+    expect(typeof fsPromises.lstat).to.equal('function');
+  });
+
+  it('should return an error if path does not exist', function() {
+    var fsPromises = util.fs().promises;
+    var fsStats;
+    fsPromises.lstat('/tmp')
```
humphd: Add a `return` statement to this line, so you return your Promise for this test.

Janice0218:
I am passed the test. @humphd

Janice0218:
I fixed the issue and passed the test! thank you! @humphd

Janice0218:
I changed one using the  git checkout master package-lock.json, is it correct? Thank you!

Janice0218:
Done! Thank you!@humphd

humphd:
Excellent, merged!

Janice0218:
Thank you for your help!! @humphd

Issue #469, opened 2018-09-24
Add test for time-flags.spec.js


sjmathew:
There are currently no tests for existing directories for filer/tests/spec/time-flags.spec.js
I can add a test case for this.

humphd:
OK @sjmathew, go for it.

PR #532, opened 2018-09-27
Fix Issue #469 - add test for time-flags with promises

sjmathew:
Test case has been added to time-flags

### DIFF
--- a/tests/spec/time-flags.spec.js
+++ b/tests/spec/time-flags.spec.js
@@ -104,3 +104,45 @@ describe('node times (atime, mtime, ctime) with mount flags', function() {
     });
   });
 });
+
+function memoryFS(flags, callback) {
+  var name = util.uniqueName();
+  return new Filer.FileSystem({
+    name: name,
+    flags: flags || [],
+    provider: new Filer.FileSystem.providers.Memory(name)
+  }, callback);
+}
+
+describe('fsPromises.update',function() {
+  beforeEach(util.setup);
+  afterEach(util.cleanup);
+  var filename = '/dir/file';
+  var newfilename = filename + '1';
+  
+  it('should not update ctime when calling fs.rename() with NOCTIME', function() {
+    memoryFS(['NOCTIME'], function(error, fs) {
+      var fsPromises = util.fs().promises;
+
+      fsPromises.createTree(fs, function() {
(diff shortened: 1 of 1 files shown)


ksandhu16:
I really like your test case well done !

tests/spec/time-flags.spec.js
```diff
@@ -104,3 +104,45 @@ describe('node times (atime, mtime, ctime) with mount flags', function() {
     });
   });
 });
+
+function memoryFS(flags, callback) {
```
humphd: When you're working on existing code in an open source project, it's important not to duplicate code.  Every time we have the same code more than once, it makes the project harder to maintain, since a bug in the code has to be fixed multiple times.  You'll hear devs say to DRY your code out, meaning "Don't Repeat Yourself."  In this case, you've copied the `memoryFS` function here, since it wasn't in the scope you wanted.  A better solution would have been to move your test into the proper scope so you could just call it.  Or, you might have refactored the method so it was available in more than one scope.

tests/spec/time-flags.spec.js
```diff
@@ -104,3 +104,45 @@ describe('node times (atime, mtime, ctime) with mount flags', function() {
...
+  }, callback);
+}
+
+describe('fsPromises.update',function() {
+  beforeEach(util.setup);
+  afterEach(util.cleanup);
+  var filename = '/dir/file';
+  var newfilename = filename + '1';
+  
+  it('should not update ctime when calling fs.rename() with NOCTIME', function() {
```
humphd: This test is passing because it's ending before any of the code executes.  Asynchronous code like this is tricky to test, because you have to make sure your test harness waits until the code in question has finished.  There are a few ways to do this.  First, we sometimes use a callback (e.g., `done` param to our function here).  You don't have one, so the test is going to be looking for your test to return a `Promise`, or it will assume that it is synchronous, and just wait til the stack finishes, and then end and assume things passed.

In this case, you need to have your top-level test function return a Promise.  You do return on line 130, but you are three call stacks deep in async code at that point, and it won't propagate back up to the right stack frame.

Make this code use the new promises stuff is going to mean doing a pretty big rewrite of `createTree` (note: it doesn't exist on `fsPromises`, even though you are calling it there.

PR #533, opened 2018-09-27
Fixes #499: Update functionality for fs.truncate

### LINKED ISSUE #499: Add a test for truncate when length is NaN

SeanPrashad:
This change to fs.truncate()'s functionality throws an error when the
length to truncate by is a non-numeric string.

### DIFF
--- a/src/filesystem/implementation.js
+++ b/src/filesystem/implementation.js
@@ -39,6 +39,7 @@ var SuperNode = require('../super-node.js');
 var Node = require('../node.js');
 var Stats = require('../stats.js');
 var Buffer = require('../buffer.js');
+const { validateInteger } = require('../shared.js');
 
 /**
  * Update node times. Only passed times are modified (undefined times are ignored)
@@ -1249,8 +1250,11 @@ function truncate_file(context, path, length, callback) {
       if(!fileData) {
         return callback(new Errors.EIO('Expected Buffer'));
       }
-      if (!Number.isInteger(length)) {
-        return callback(new Errors.EINVAL('length must be a number'));
+      try {
+        validateInteger(length, 'len');
+      }
+      catch (error) {
+        return callback(error);
       }
       var data = new Buffer(length);
       data.fill(0);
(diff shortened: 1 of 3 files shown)


src/filesystem/implementation.js
```diff
@@ -1249,6 +1249,9 @@ function truncate_file(context, path, length, callback) {
       if(!fileData) {
         return callback(new Errors.EIO('Expected Buffer'));
       }
+      if (!Number.isInteger(length)) {
```
humphd: node does this: https://github.com/nodejs/node/blob/master/lib/internal/validators.js#L62-L76  Do you want to just add that function in here and call it?  There might be a few other places that need it.

SeanPrashad: Great point - do you think we should put these validation functions into a separate file so that other parts of the codebase can use it?

humphd: Throw it in `shared.js`

src/shared.js
```diff
@@ -19,8 +21,24 @@ function u8toArray(u8) {
...
 
+function validateInteger(value, name) {
+  let err;
+
+  if (typeof value !== 'number')
+    err = new Errors.EINVAL(name, 'number', value);
+
+  if (err) {
+    Error.captureStackTrace(err, validateInteger);
+    throw err;
```
SeanPrashad: I kept the throw instead of returning the error to keep it inline with how Node.js does it - what do you think?

humphd: Yeah, match node's API and behaviour.

Issue #473, opened 2018-09-24
Adding test for fs.writeFile() when the string is externalized.


ksandhu16:
I noticed there was no test case for fs.writeFile(). I would like to add a test case for externalized string handling. I would like to be assigned to this issue.

PR #539, opened 2018-09-28
fix 473 for fs.WriteFile 

ksandhu16:


VictorKubrak:
This test seems to already exist: https://github.com/filerjs/filer/blob/master/tests/spec/fs.writeFile-readFile.spec.js
But if you still want to write this test, try adding data parameter to your writeFile().
Look https://nodejs.org/api/fs.html#fs_fs_writefile_file_data_options_callback for method description.

tests/spec/fs.writeFile-readFile.spec.js
```diff
@@ -99,3 +99,25 @@ describe('fs.writeFile, fs.readFile', function() {
     });
   });
 });
+
+describe('fsPromises.Write', function() {
+  it('should write, read a utf8 file without specifying utf8 in writeFile', function() {
+    //var fs = util.fs();
+    //var contents = 'This is a file.';
```
humphd: You don't want to leave commented code in a PR like this.  If you don't need the code, remove it.  There's no need to keep it around, since we can go back in time with git to recover old code if necessary.

tests/spec/fs.writeFile-readFile.spec.js
```diff
@@ -99,3 +99,25 @@ describe('fs.writeFile, fs.readFile', function() {
...
+	
+    var fsPromises = util.fs().promises;
+	
+    return fsPromises.writeFile('/myfile' , { encoding: 'utf8' })
```
humphd: `writeFile` needs an argument with the `data` to write

tests/spec/fs.writeFile-readFile.spec.js
```diff
@@ -99,3 +99,25 @@ describe('fs.writeFile, fs.readFile', function() {
...
+  it('should write, read a utf8 file without specifying utf8 in writeFile', function() {
+    //var fs = util.fs();
+    //var contents = 'This is a file.';
+	
+    var fsPromises = util.fs().promises;
+	
+    return fsPromises.writeFile('/myfile' , { encoding: 'utf8' }) 
+      
+      .then(() => {
+        return  fsPromises.readFile('/myfile', 'utf8')
```
humphd: You can shorten this to:

```js
.then(() => fsPromises.readFile('/myfile', 'utf8'))
.then(data => {
  expect(data).to.equal(contents); // here assuming you have some kind of data contents defined above
});
```

tests/spec/fs.writeFile-readFile.spec.js
```diff
@@ -99,3 +99,25 @@ describe('fs.writeFile, fs.readFile', function() {
...
+          
+          .catch((error) => {
```
humphd: This isn't necessary

tests/spec/fs.writeFile-readFile.spec.js
```diff
@@ -99,3 +99,25 @@ describe('fs.writeFile, fs.readFile', function() {
...
+		
+            expect(error).not.to.exist;
+            //expect(data).to.equal(contents);
```
humphd: Not sure why these are commented out?

tests/spec/fs.writeFile-readFile.spec.js
```diff
@@ -99,3 +99,25 @@ describe('fs.writeFile, fs.readFile', function() {
...
+            //done();
+          });
+      });
+  });
+});
```
humphd: Always include a newline at the end of each file.

humphd:
NOTE: when you create a PR, make sure you provide both a Subject line and a full Description to describe what the changes do, why you did what you did, which Issue # it fixes (e.g., Fixes #473).

Issue #527, opened 2018-09-26
Add tests for fs.promises utimes function 


biskit1:
Perform the existing tests for the utimes function but with fs.promises.utimes() to ensure that the integration of promises in filer works as expected.
In addition, check that fs.promises.utimes() returns a promise as expected. 

Can I work on this issue?

SeanPrashad:
Hey @biskit1 - how are you doing on this?

biskit1:
Thanks for checking in @SeanPrashad; should have this ready for review by end of day.

biskit1:
https://github.com/filerjs/filer/blob/e3a1187ef94d5fc627768cee13457417e7ea0035/tests/spec/fs.utimes.spec.js#L86

Is it a mistake that `futimes` is called instead of `utimes` @humphd? There are no tests for `futimes` version of the function.

modeswitch:
@biskit1 The tests for the file descriptor versions of each function (`fstat`, `futimes`, etc) live with the non-descriptor versions. That particular test is making sure that passing a bad descriptor to `futimes` has the correct behavior. The test only makes sense for `futimes` since `utimes` doesn't accept a descriptor.

biskit1:
Thanks @modeswitch, that makes sense. Should have spent more time tracking the calls. That would also probably explain why my test is failing (https://github.com/filerjs/filer/pull/540/files#r221447167- except I still don't get why it was passing when I did not have the `return` there- if you see this and have an answer before I come up with one, please share 😏).

PR #540, opened 2018-09-30
Add tests for fs.promises utimes function #527

biskit1:
Fixes issue #527 
Tests if `utimes` function returns a promise and implements promise versions of existing tests.

### DIFF
--- a/tests/spec/fs.utimes.spec.js
+++ b/tests/spec/fs.utimes.spec.js
@@ -242,7 +242,7 @@ describe('fs.promises.utimes', function () {
     var atime = Date.parse('1 Oct 2000 15:33:22');
     var mtime = Date.parse('30 Sep 2000 06:43:54');
 
-    fs.utimes(1, atime, mtime)
+    return fs.utimes(1, atime, mtime)
       .catch(function (error) {
         expect(error).to.exist;
         expect(error.code).to.equal('EBADF');
@@ -254,7 +254,7 @@ describe('fs.promises.utimes', function () {
     var atime = Date.parse('1 Oct 2000 15:33:22');
     var mtime = Date.parse('30 Sep 2000 06:43:54');
 
-    fs.writeFile('/testfile', '')
+    return fs.writeFile('/testfile', '')
       .then(function () {
         fs.utimes('/testfile', atime, mtime)
           .then(function () {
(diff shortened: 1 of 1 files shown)


tests/spec/fs.utimes.spec.js
```diff
@@ -175,3 +175,186 @@ describe('fs.utimes', function() {
...
+    fs.writeFile('/testfile', '')
+      .then(function () {
+        fs.utimes('/testfile', atime, mtime)
+          .then(function () {
+            fs.stat('/testfile')
+              .then(function (stat) {
+                expect(stat.mtime).to.equal(mtime);
+              })
+              .catch(function (error) {
+                expect(error).not.to.exist;
```
biskit1: Wondering if the `expect(error)` and entire `.catch()` portion is necessary for testing? 
If I left it out, what would happen if `.then()` failed?

tests/spec/fs.utimes.spec.js
```diff
@@ -242,7 +242,7 @@ describe('fs.promises.utimes', function () {
     var atime = Date.parse('1 Oct 2000 15:33:22');
     var mtime = Date.parse('30 Sep 2000 06:43:54');
 
-    fs.utimes(1, atime, mtime)
+    return fs.utimes(1, atime, mtime)
```
biskit1: Not sure why all the tests passed when I forgot the return statements, but now that I added them just this one test is failing

biskit1: Should have called `futimes`, that's why it failed.

biskit1: @humphd, why do all the tests pass when I leave out the `return` statements?

humphd: The test hits the end of the stack, and nothing failed, so it says, "OK, I guess this worked."  Meanwhile, your file system calls are running async, and haven't started (or finished) yet:

```js
it('...', function() {
  startSomethingAsync();
  //test will end right here, before startSomethingAsync can finish
});
```

However, when you `return` a promise (or with callbacks use `done`), you block the test from exiting until the task is complete (promise is resolved, or callback done() is executed).

biskit1: thank you!

Issue #542, opened 2018-10-04
Add test for fsPromises.symlink(target, path[, type]) to return error if part of parent destination path doesn't exist


deepanjali19:
I notice that there is currently no test for [fsPromises.symlink(target, path[, type])](https://nodejs.org/api/fs.html#fs_fspromises_symlink_target_path_type).

I would like to work on the test method to check any errors if [part of the parent destination path does not exist](https://github.com/filerjs/filer/blob/master/tests/spec/fs.symlink.spec.js#L13).

I will be putting up the pull request soon!

PR #543, opened 2018-10-04
fix for #542 for fsPromises.symlink(target, path[, type])

deepanjali19:
This is the fix for issue #542

Now there is a test for `fsPromises.symlink(target, path[, type])` to return error if part of the parent destination path doesn't exist.


tests/spec/fs.symlink.spec.js
```diff
@@ -44,3 +44,21 @@ describe('fs.symlink', function() {
...
+describe('fsPromises.symlink', function () {
+  beforeEach(util.setup);
+  afterEach(util.cleanup);
+
+
+  it('should return an error if part of the parent destination path does not exist', () => {
+    var fsPromises = util.fs().promises;
+
+    return fsPromises.symlink('/', '/tmp/mydir')
+      .then()
```
humphd: Remove this, since it's not doing anything.

tests/spec/fs.symlink.spec.js
```diff
@@ -44,3 +44,21 @@ describe('fs.symlink', function() {
...
+      .catch(error => {
+        expect(error).to.exist;
+        expect(error.code).to.equal('ENOENT');
+      });
+  });
+});
```
humphd: Newline at the end of each file please.

PR #546, opened 2018-10-09
Fix appendFile bug shown to fail in PR-494

### LINKED ISSUE #463: add tests for fs.appendFile for valid file system flags

MuchtarSalimov:
As seen in PR #494, the current implementation of appendFile fails when passing a parameter that looks like { flag: 'a' } (which should succeed, since 'a' is a default value). I have included a fix and the accompanying test to verify that it works (that previous failed with the current implementation).

The reason it failed on this type of input is that validate_file_options in [implementation.js](https://github.com/MuchtarSalimov/filer/blob/master/src/filesystem/implementation.js) only checks if the parameter is null, a function, or a string. If it is something else (which in this case in an object), it just returns it as-is. But in the case of having an object but not having an encoding specified (as in { flags: 'a' }),  options.encoding is not set to the default value.

Fixes #463

### DIFF
--- a/src/filesystem/implementation.js
+++ b/src/filesystem/implementation.js
@@ -1870,12 +1870,17 @@ function appendFile(fs, context, path, data, options, callback) {
   if(!flags) {
     return callback(new Errors.EINVAL('flags is not valid', path));
   }
+  
+  if (typeof options === 'object' && options.encoding === undefined ) {
+    options.encoding = 'utf8';
+  }
 
   data = data || '';
   if(typeof data === 'number') {
     data = '' + data;
   }
-  if(typeof data === 'string' && (options.encoding === null || options.encoding === 'utf8')) {
+  
+  if(typeof data === 'string' && options.encoding === 'utf8') {
     data = Encoding.encode(data);
   }
 
--- a/src/filesystem/implementation.js
+++ b/src/filesystem/implementation.js
@@ -1875,7 +1875,7 @@ function appendFile(fs, context, path, data, options, callback) {
   if(typeof data === 'number') {
     data = '' + data;
   }
-  if(typeof data === 'string' && (options.encoding == null || options.encoding === 'utf8')) {
+  if(typeof data === 'string' && (options.encoding === null || options.encoding === 'utf8')) {
     data = Encoding.encode(data);
   }
 
--- a/src/filesystem/implementation.js
+++ b/src/filesystem/implementation.js
@@ -1874,6 +1874,18 @@ function appendFile(fs, context, path, data, options, callback) {
   if (typeof options === 'object' && options.encoding === undefined ) {
     options.encoding = 'utf8';
   }
+  
+  if typeof (options ==== 'object') {
+    if (options.encoding === undefined) {
+      options.encoding = 'utf8';
+    }
+    if (options.mode === undefined) {
+      options.mode = 0o666;
+    }
+    if (options.flag === undefined) {
+      options.flag = 'a';
+    }
+  }
 
   data = data || '';
   if(typeof data === 'number') {
--- a/src/filesystem/implementation.js
+++ b/src/filesystem/implementation.js
@@ -1871,7 +1871,7 @@ function appendFile(fs, context, path, data, options, callback) {
     return callback(new Errors.EINVAL('flags is not valid', path));
   }
   
-  if (typeof options ==== 'object') {
+  if (typeof options === 'object') {
     if (options.encoding === undefined) {
       options.encoding = 'utf8';
     }
--- a/src/filesystem/implementation.js
+++ b/src/filesystem/implementation.js
@@ -1871,10 +1871,6 @@ function appendFile(fs, context, path, data, options, callback) {
     return callback(new Errors.EINVAL('flags is not valid', path));
   }
   
-  if (typeof options === 'object' && options.encoding === undefined ) {
-    options.encoding = 'utf8';
-  }
-  
   if typeof (options ==== 'object') {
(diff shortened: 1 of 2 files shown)


humphd (changes requested):
This is great that you've also gone ahead and fixed the issue you found with the tests, nice work!

Because you've built on the same set of tests, and included them here too, I'm going to close #494 and use this PR instead.  I'll obviously still count that for your 0.1 release, but it will make landing this fix easier.

humphd:
@MuchtarSalimov that broke your test: `  ✖ should append without error when specfifying flag option (default value)`

src/filesystem/implementation.js
```diff
@@ -1870,11 +1870,24 @@ function appendFile(fs, context, path, data, options, callback) {
   if(!flags) {
     return callback(new Errors.EINVAL('flags is not valid', path));
   }
+  
+  if (typeof options === 'object') {
+    if (options.encoding === undefined) {
```
humphd: Prefer a check for `!options.*` for all of these, so we also deal with things that aren't explicitly `undefined`

```suggestion
    if (!options.encoding) {
```

humphd (changes requested):
Looks good, just switch those `if` checks to not use `undefined.

src/filesystem/implementation.js
```diff
@@ -1870,11 +1870,24 @@ function appendFile(fs, context, path, data, options, callback) {
   if(!flags) {
     return callback(new Errors.EINVAL('flags is not valid', path));
   }
+  
+  if (typeof options === 'object') {
```
humphd: I'm confused about what you're doing here, because `validate_file_options` above on line 1865 does the same thing, see https://github.com/filerjs/filer/blob/8c7d873c4c2497af73d60b12719a5a73e1eed2f6/src/filesystem/implementation.js#L1575-L1584

The only thing it's missing is the `mode`.  By the time we get to your code here, `options` should always already be an `Object` with `encoding` and `flag`, so your code here is essentially not doing anything.

Should we just move your `options.mode` up to 1866 like so:

```js
options.mode = options.mode || 0o666;
```

humphd (changes requested):

src/filesystem/implementation.js
MuchtarSalimov: I think you're right about setting options.flag and options.mode not doing anything, since not only does it come after line 1869, that line also takes care of it being undefined as well.
`var flags = validate_flags(options.flag || 'a');`
As for mode, I don't see it being called at all.


However,

1) options needs not be an object by this point. If it is integer, it never gets replaced with an object, and checking options.encoding then seems wrong to me. That's why I check for the object type.

2) The options object needs not be fully formed. If the object passed is empty:
`{}`
 or missing options.encoding:
`{ flag = 'a' } `
it breaks because validate_file_options _assumes_ that it is fine and leaves it alone.


If you would like, I can:

- remove the extra lines related to options.flag
- remove the extra lines related to options.mode (or change it to 1866, but I don't see it being used anywhere)

humphd: According to https://nodejs.org/api/fs.html#fs_fs_appendfile_path_data_options_callback, `options` has to be one of a) an `Object`; b) a `String` and *"If options is a string, then it specifies the encoding"*

This is why `validate_file_options` is checking for the cases:

1. no `options` arg given at all, in which case it creates a default `Object` that's only missing the `mode` default value
2. `options` is a `function, which is a special case of 1., where not even args were passed to the `appendFile`, and we didn't get an `options` `Object`, only the `callback`.
3. `options` is a `String`, which means we've been passed the `encoding` value, and we package that up as an `Object.

All of that to say, it can't be an `Integer` at this point, it has to be an `Object`.  If you try and run this in node (i.e., with `options` as a `Number`), here is what happens:

```
> require('fs').appendFile('./myfile', 'data', 1, err => console.log(err));
TypeError: "options" must be a string or an object, got number instead.
    at getOptions (fs.js:75:11)
    at Object.fs.appendFile (fs.js:1323:13)
```

So maybe we should also add a test to make sure that if you pass a `Number`, it will fail the way we expect.

However, I agree that it never checks that `options` as an `Object` has the correct shape (properties we expect).  Out of interest, here is what node does:

https://github.com/nodejs/node/blob/eef072fa083f05f84fa6ca1908472eb228095a38/lib/fs.js#L1227-L1239

Which is calling `getOptions`:

https://github.com/nodejs/node/blob/eef072fa083f05f84fa6ca1908472eb228095a38/lib/internal/fs/utils.js#L165-L182

I think you should do the following:

1. Call `validate_file_options` to try and get `options` in the right form
2. Do a check similar to what node does with what you get back (i.e., make sure it's an `Object`:
    ``js
    if (typeof options !== 'object') {
        throw new ERR_INVALID_ARG_TYPE('options', ['string', 'Object'], options);
    }
    ```
3. Make sure that `options` has the correct shape (all properties are there, or set defaults).

This is pretty close to what you have now, with the exception that you're silently allowing non-`Object` types to get through.

Issue #493, opened 2018-09-24
Add test for rename() when oldPath and newPath are the same


VictorKubrak:
I found out that existing tests do not test this case. Function should not rename the file, then it returns success status.
I would like to add this test myself.

humphd:
OK @VictorKubrak, go for it.

Pooch11:
By the looks of things in other issues - the goal is to move closer to IEEE standard.
 Consider looking at 
[The Open Group](http://pubs.opengroup.org/onlinepubs/9699919799/)
this will show what the expected behaviour of this function is. I did not see any restrictions on renaming a file with the same name. Nevertheless - its probably a missing test!

PR #552, opened 2018-10-14
Fixed #548 for rename()

### LINKED ISSUE #548: rename() fails to rename a file to itself
I was writing a test for rename() when oldPath==newPath (You can see a discussion [here](https://github.com/filerjs/filer/pull/502)). It was expected that function will successfully rename a file, but it fails. We came to the conclusion that function implementation has a bug.
I would like this bug to be assigned to me.

VictorKubrak:
- Fixed [issue-548](https://github.com/filerjs/filer/issues/548): rename() failed to rename a file to the same filename
- Also added a test that solves [issue-493](https://github.com/filerjs/filer/issues/493). It tests case when oldPath==newPath
- So, this PR can be merged with [pull-502](https://github.com/filerjs/filer/pull/502)

### DIFF
--- a/src/filesystem/implementation.js
+++ b/src/filesystem/implementation.js
@@ -933,13 +933,9 @@ function link_node(context, oldpath, newpath, callback) {
       callback(error);
     } else {
       newDirectoryData = result;
-      if(_(newDirectoryData).has(newname)) {
-        callback(new Errors.EEXIST('newpath resolves to an existing file', newname));
-      } else {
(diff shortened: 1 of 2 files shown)


modeswitch:
I'm reviewing this now. Sorry for the huge delay. I need a sec to catch up on all the context as well.

humphd:
@modeswitch can you bless this, and I'll rebase and land.  I think it's good.

PR #553, opened 2018-10-17
Issue 551-Add unwatchfile method

VinceBackpack:
This method works for the first test case from [issue-445](https://github.com/filerjs/filer/issues/445)
but for the second method it gives typeerror: this.off is not a function.

Can someone help point me to the right direction.

### DIFF
--- a/src/filesystem/interface.js
+++ b/src/filesystem/interface.js
@@ -17,6 +17,7 @@ var providers = require('../providers/index.js');
 var Shell = require('../shell/shell.js');
 var Intercom = require('../../lib/intercom.js');
 var FSWatcher = require('../fs-watcher.js');
+//var unwatcher = require('../unwatcherFile.js');
 var Errors = require('../errors.js');
 var defaultGuidFn = require('../shared.js').guid;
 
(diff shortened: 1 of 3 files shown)


src/filesystem/interface.js
```diff
@@ -159,6 +160,27 @@ function FileSystem(options, callback) {
...
+    if(isNullPath(filename)) {
+      throw new Error('Path must be a string without null bytes.');
+    }
+    listener = listener || nop;
+
+    if(listener == nop){
+      this.removeAllListeners();
+    }
+    else{
+      this.off('change', listener);
```
humphd: For what you're trying to do to work, we're going to have to alter how `.watch()` works above, namely, we're going to need to cache the listener for the given filename on an `Object` accessible by both of these functions, and keyed on the filename, such that we can access it here.

Probably on line 159 above, when you have both the `filename`, the `watcher`, and the `listener`, you'll need to add it to an `Object` that we define somewhere, either in this file on line 143, or maybe we do it in `FSWatcher` itself.

Probably the latter is best, since we're normalizing the filename there.  Over [here](https://github.com/filerjs/filer/blob/db315eae134c1e0d9d9b47531615b13c4abb9026/src/fs-watcher.js#L61) you need something like this:

```js
// At the top of the file
var _watchers = {};

FSWatcher.getWatcherForFilename = function(filename) {
  return _watchers[filename];
};
```

Now in the code for `start()`, you can cache the `watcher` instance in `_watchers`, maybe [here](https://github.com/filerjs/filer/blob/db315eae134c1e0d9d9b47531615b13c4abb9026/src/fs-watcher.js#L48):

```js
_watchers[filename] = self;
```
Now you have what you need to call `.close()` here when you want to remove this watcher.

Do you follow what I'm saying?

PR #560, opened 2018-11-15
Fix #558 - Implementing fs.access() method

### LINKED ISSUE #558: Implement fs.access() method
@deepanjali19 noticed that we don't have any implementation for `fs.access()`:

* [node docs for `fs.access`](https://nodejs.org/api/fs.html#fs_fs_access_path_mode_callback)
* [code in node.js](https://github.com/nodejs/node/blob/af6d26281f734dbbf1bf497103d1cb4b85a52b2e/lib/fs.js#L174)

The general pattern for implementing a new method like this in Filer is to do the following:

1. Add the new `access` method name to the [interface list](https://github.com/filerjs/filer/blob/master/src/filesystem/interface.js#L271-L305)
1. Next add a new exported method name for `access` to [this list in the implementation file](https://github.com/filerjs/filer/blob/master/src/filesystem/implementation.js#L2386L2420).
1. Now add a new file to the tests: `tests/spec/fs.access.js`.  Add a very simple set of tests to start (you can add more later below), something like this:

    ```js
    var util = require('../lib/test-utils.js');

    describe('fs.chmod, fs.fchmod', function() {
       beforeEach(util.setup);
       afterEach(util.cleanup);

       it('should be functions', function() {
           var fs = util.fs();
           expect(typeof fs.access).to.equal('function');
       });
    });
    ````
1. Add your new test file `tests/spec/fs.access.js` to the list of [`required()` test files here](https://github.com/filerjs/filer/blob/master/tests/index.js#L10-L44).
1. Write an implementation for the `fs.access` method in `implementation.js` that looks something like this:
    ```js
    function access(fs, context, path, mode, callback) {
        //...
    }
    ```
    You need to do a few things here.  You'll have to deal with the `mode` not being given, and instead getting `callback` on `mode`.  You can look at various other functions in this file which deal with the same problem, for example [`mkdir`](https://github.com/filerjs/filer/blob/master/src/filesystem/implementation.js#L1675-L1686)
1. Next you'll need to write a function `access_file` which gets called by your `access` function.  Basically, the `access` function is the public API, and here you need to deal with all the various arguments being passed in (e.g., set defaults, check that things make sense).  Then, the bulk of the work happens in `access_file`.  In our case, we'll need to get the filesystem `Node` for the given path, get it's `mode`, and then do bitwise comparison to see if the `mode` flag passed to `access` matches.  I can help you more when you get to this stage.
1. Finally, we'll need to write all the tests for every possible case of calling `fs.access` with different paths, modes, etc.

Assigning to @deepanjali19, who offered to work on this.  Please let me know if you have questions, and I can help guide you.

deepanjali19:
I have created the implementation for `fs.access()`.

At this point it only supports testing for `fs.constants.F_OK`. I will be adding other use cases soon.


tests/spec/fs.access.js
```diff
@@ -0,0 +1,38 @@
...
+
+    fs.access('/tmp', fs.constants.F_OK, function (error) {
+      expect(error).to.exist;
+      expect(error.code).to.equal('ENOENT');
+      done();
+    });
+  });
+
+  it('should return no error if file does exist', function (done) {
+
```
humphd: Remove this empty line

src/filesystem/implementation.js
```diff
@@ -48,28 +48,28 @@ const { validateInteger } = require('../shared.js');
 function update_node_times(context, path, node, times, callback) {
   // Honour mount flags for how we update times
   var flags = context.flags;
-  if(_(flags).contains(FS_NOCTIME)) {
+  if (_(flags).contains(FS_NOCTIME)) {
```
humphd: It looks like your editor has done a ton of formatting changes in this file.  We should undo that, so we don't lose all our git history.

My suggestion to you is to do the following:

1. copy/paste your changes from this file so you don't lose them.
1. `git checkout master src/filesystem/implementation.js`
1. re-add your copy/pasted changes from `src/filesystem/implementation.js` back to the file
1. add, commit, push this back up.

src/filesystem/implementation.js
```diff
@@ -1678,23 +1693,36 @@ function mkdir(fs, context, path, mode, callback) {
...
   }
- 
-  if(!pathCheck(path, callback)) return;
+
+  if (!pathCheck(path, callback)) return;
   make_directory(context, path, callback);
 }
 
+function access(fs, context, path, mode, callback) {
+
```
humphd: Remove this blank line

src/filesystem/implementation.js
```diff
@@ -441,6 +441,21 @@ function make_directory(context, path, callback) {
   find_node(context, path, check_if_directory_exists);
 }
 
+/**
```
humphd: You can remove this comment, as it doesn't add anything

src/filesystem/implementation.js
```diff
@@ -441,6 +441,21 @@ function make_directory(context, path, callback) {
...
+ * access_file
+ */
+function access_file(context, path, mode, callback) {
+  path = normalize(path);
+  find_node(context, path, function (err) {
+    if (err) {
+      return callback(err);
+    }
+    // TO DO: we currently only support Constants.fsConstants.F_OK
```
humphd: Update this comment to also include a reference to the Issue I filed https://github.com/filerjs/filer/issues/561.

nit: change `TO DO` to `TODO` (no space)

humphd:
One more thing we're going to need is to have the `README.md` file updated to include docs for `fs.access`.  You can follow the same pattern as the other methods, just do a search in that file for something like `mkdir` and notice all the places and ways it's referenced in the markdown.

deepanjali19:
Working in README.md 👍

humphd:
Filed https://github.com/filerjs/filer/issues/562 on the test failure, it's not related to your changes.  Is this ready for review again?

src/filesystem/implementation.js
```diff
@@ -1250,12 +1263,6 @@ function truncate_file(context, path, length, callback) {
       if(!fileData) {
         return callback(new Errors.EIO('Expected Buffer'));
       }
-      try {
```
humphd: How come this got removed?  I wonder if this is your bug in Travis.

src/filesystem/implementation.js
```diff
@@ -39,7 +39,6 @@ var SuperNode = require('../super-node.js');
 var Node = require('../node.js');
 var Stats = require('../stats.js');
 var Buffer = require('../buffer.js');
-const { validateInteger } = require('../shared.js');
```
humphd: Same here, this can go back in.

deepanjali19:
@humphd I am almost done except 1 thing. I am not sure about **FileSystemShell Instance Method for fs.access**. I would really appreciate if you can give me some knowledge about it. 
Secondly, do you want me to make any changes regarding **fs.truncate**?

src/filesystem/implementation.js
deepanjali19: I am almost sure I did not touch that part.

humphd: OK, it needs to go back in.  Would you be able to add back this section and the other on line 42?  That will fix that error.

deepanjali19:
@humphd ready for the review.

README.md
```diff
@@ -672,6 +673,29 @@ fs.mkdir('/home', function(err) {
...
+#### fs.access(path, [mode], callback)<a name="access"></a>
+
+Tests a user's permissions for the file or directory supplied in `path` argument. Asynchronous [access(2)](http://pubs.opengroup.org/onlinepubs/009695399/functions/access.html). Callback gets no additional arguments. The `mode` argument can be:
+
+* `F_OK`: Test for existence of file.
+* `R_OK`: Test whether the file exists and grants read permission.
+* `W_OK`: Test whether the file exists and grants write permission.
+* `X_OK`: Test whether the file exists and grants execute permission.
+
+NOTE: Filer allows for, but ignores the optional `mode` argument used in node.js.
```
humphd: Remove this.

README.md
```diff
@@ -672,6 +673,29 @@ fs.mkdir('/home', function(err) {
 });
 ```
 
+#### fs.access(path, [mode], callback)<a name="access"></a>
+
+Tests a user's permissions for the file or directory supplied in `path` argument. Asynchronous [access(2)](http://pubs.opengroup.org/onlinepubs/009695399/functions/access.html). Callback gets no additional arguments. The `mode` argument can be:
```
humphd: "...argument can be one of the following (constants are available on `fs.constants`):"

README.md
humphd: Instead, can you add  this:

"NOTE: you can also create a mask consisting of the bitwise OR of two or more values (e.g. `fs.constants.W_OK | fs.constants.R_OK`)."

README.md
```diff
@@ -672,6 +673,29 @@ fs.mkdir('/home', function(err) {
...
+NOTE: Filer allows for, but ignores the optional `mode` argument used in node.js.
+
+Example:
+
+```javascript
+/*
+* Given the following file exists:
+* '/myfile'
+*/
+ fs.access('/myfile', fs.constants.F_OK, function (error) {
```
humphd: Why don't we steal the docs from node here instead:

```
// Check if the file exists in the current directory.
fs.access(file, fs.constants.F_OK, function(err) {
  console.log(`${file} ${err ? 'does not exist' : 'exists'}`);
});
```

It's a nice example showing how to interpret the presence of `err`

tests/index.js
```diff
@@ -9,6 +9,7 @@ require('./spec/filer.spec');
 // Filer.FileSystem.*
 require('./spec/filer.filesystem.spec');
 require('./spec/fs.spec');
+require('./spec/fs.access');
```
humphd: Can you rename your file `tests/spec/fs.access.spec.js` to match everything else and update this line here to also include `.spec` at the end?

README.md
```diff
@@ -672,6 +673,26 @@ fs.mkdir('/home', function(err) {
 });
 ```
 
+#### fs.access(path, [mode], callback)<a name="access"></a>
+
+Tests a user's permissions for the file or directory supplied in `path` argument. Asynchronous [access(2)](http://pubs.opengroup.org/onlinepubs/009695399/functions/access.html). Callback gets no additional arguments. The `mode` argument can be one of the following (constants are available on fs.constants):
```
humphd: Wrap fs.constants in backticks please.

README.md
```diff
@@ -672,6 +673,26 @@ fs.mkdir('/home', function(err) {
...
+
+* `F_OK`: Test for existence of file.
+* `R_OK`: Test whether the file exists and grants read permission.
+* `W_OK`: Test whether the file exists and grants write permission.
+* `X_OK`: Test whether the file exists and grants execute permission.
+
+NOTE: you can also create a mask consisting of the bitwise OR of two or more values (e.g. fs.constants.W_OK | fs.constants.R_OK).
```
humphd: Same thing here, wrap fs.constants.W_OK | fs.constants.R_OK in backticks.

PR #564, opened 2018-11-17
Added fs.promises.write test

susantruong:
Added `fs.promises.write` test


humphd:
The code on master has evolved since this test was written, and the cases it adds are no longer valid (i.e., `fs.write` doesn't return a promise).  Thanks for starting this, and sorry I have to close it.

PR #565, opened 2018-11-18
Fixes #422: Add nyc lcov-based code coverage for tests

### LINKED ISSUE #422: Handling of constant values in Filer's implementation
As I was traversing through the [implementation.js](https://github.com/filerjs/filer/blob/fd3de6be2c1c19421d8c83e6791cd4b19afb84c7/src/filesystem/implementation.js#L22) file, I realized that something like `var O_READ = Constants.O_READ` is trying to create a constant variable. However, you could later set `O_READ = Constant.O_WRITE` if you so pleased...

Would it be better in this scenario to use `const` instead of `var` ? This would ensure that these values are not changed by mistake in other function calls (for instance). 

Again, I can make the hot-fix for this issue if need be.

humphd:
This adds code coverage support to our tests.  I had originally wanted to do this via Karma when we run our main tests.  The way that we do our bundle, I can't figure out a good solution for getting accurate source line info.

To get around this, I've added a separate test run called `coverage`, which uses [nyc](https://github.com/istanbuljs/nyc) with mocha in node.js to do the tests, and instruments the code (unbundled) with Istanbul.  This produces lcov reports, as well as text and HTML.

I had to adjust a few things in the tests to get the working well in node.js.  The only thing this doesn't instrument is any of our indexeddb or websql provider code.  That's not a huge problem, because we use it 100% in the browser-based Karma tests.  I was more interested to see how things in `src/` were doing.

I'll attach the report in a sec, but the good news is we're doing really well with test coverage.

@modeswitch one question I have: do you want me to hook this up to https://codecov.io or https://coveralls.io/ with Travis?  What it would do is give us per-commit reports on coverage.  Here's an example of what the reports look like on CodeCov:

https://codecov.io/gh/mozilla/addons-frontend/src/master/src/core/client/base.js

[It's easy to do in this PR or later](https://github.com/istanbuljs/nyc#integrating-with-coveralls), if we choose.  Let me know what you think.

cc'ing @MarkKrutik who started this for us in #464, which I've closed in favour of this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,7 @@ dist/filer-issue225.js
 # Parcel build dirs
 .cache
 tests/dist
+
+# nyc code coverage
+.nyc_output
+coverage
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,6 +4,160 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@babel/code-frame": {
+      "version": "7.0.0",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.0.0.tgz",
+      "integrity": "sha512-OfC2uemaknXr87bdLUkWog7nYuliM9Ij5HUcajsVcMCpQrcLmtxRbVFTIqmcSkSeYRBFBRxs2FiUqFJDLdiebA==",
+      "dev": true,
+      "requires": {
+        "@babel/highlight": "^7.0.0"
+      }
+    },
+    "@babel/generator": {
+      "version": "7.1.6",
+      "resolved": "https://registry.npmjs.org/@babel/generator/-/generator-7.1.6.tgz",
+      "integrity": "sha512-brwPBtVvdYdGxtenbQgfCdDPmtkmUBZPjUoK5SXJEBuHaA5BCubh9ly65fzXz7R6o5rA76Rs22ES8Z+HCc0YIQ==",
+      "dev": true,
+      "requires": {
+        "@babel/types": "^7.1.6",
+        "jsesc": "^2.5.1",
+        "lodash": "^4.17.10",
+        "source-map": "^0.5.0",
+        "trim-right": "^1.0.1"
+      },
+      "dependencies": {
+        "jsesc": {
+          "version": "2.5.2",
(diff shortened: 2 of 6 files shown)


humphd:
Here's what our current test coverage looks like:

```
---------------------------|----------|----------|----------|----------|-------------------|
File                       |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
---------------------------|----------|----------|----------|----------|-------------------|
All files                  |    85.57 |    74.18 |    96.52 |     85.8 |                   |
 src                       |    92.09 |    82.99 |    96.23 |    92.34 |                   |
  buffer.js                |      100 |       75 |      100 |      100 |                18 |
  constants.js             |      100 |      100 |      100 |      100 |                   |
  directory-entry.js       |      100 |       50 |      100 |      100 |                 5 |
  encoding.js              |      100 |      100 |      100 |      100 |                   |
  errors.js                |      100 |      100 |      100 |      100 |                   |
  fs-watcher.js            |    92.31 |    76.92 |      100 |    92.31 |             27,31 |
  index.js                 |      100 |      100 |      100 |      100 |                   |
  node.js                  |    91.11 |     92.5 |      100 |    91.11 |       64,65,70,71 |
  open-file-description.js |    93.33 |       75 |      100 |    93.33 |                18 |
  path.js                  |    90.91 |    75.81 |    94.12 |    91.76 |... 02,169,184,191 |
  shared.js                |    70.59 |      100 |       80 |    68.75 |    16,17,18,19,21 |
  stats.js                 |      100 |      100 |      100 |      100 |                   |
  super-node.js            |     87.5 |       90 |      100 |     87.5 |             18,19 |
 src/filesystem            |    83.71 |    71.96 |    99.14 |    83.91 |                   |
  implementation.js        |    82.54 |    71.88 |      100 |    82.73 |... 2308,2316,2378 |
  interface.js             |    92.62 |    73.08 |    92.86 |    92.62 |... 77,184,232,257 |
 src/shell                 |    87.69 |    76.71 |    86.67 |    87.97 |                   |
  environment.js           |      100 |      100 |      100 |      100 |                   |
  shell.js                 |    87.26 |    75.71 |    85.96 |    87.55 |... 03,520,528,529 |
---------------------------|----------|----------|----------|----------|-------------------|
```

Most of what isn't covered is stuff we don't really care about (checks on missing callbacks dealt with elsewhere, for example).

PR #567, opened 2018-11-19
Fix #566: support both node.mode and node.posix_mode

### LINKED ISSUE #566: Prevent data loss on Filer upgrade, deal with changes to Node layout
In https://github.com/filerjs/filer/commit/c526445a4358f74865f7bcbdcf0a43514940bf4d and https://github.com/filerjs/filer/commit/ee412d4abe259a08ba036cd6d980f469ff8a2ef0 I made some adjustments to the layout of our filesystem `Node`.  Specifically, I changed the meaning of `mode` from file type to be a POSIX style mode number.

As we get ready to do fix #555, I'm realizing that I don't want to break existing filesystems when people upgrade to this new code.

We need a way for a current Node layout to get upgraded to the new layout in place, without data loss.

Probably the easiest thing to do here is for me to switch `type` back to `mode`, and introduce a new property called `posix_mode` or something.

@modeswitch can you help me think this through before you land that?  I can do the PR.

humphd:
I've taken a stab at fixing this so existing filesystems won't break when Filer is upgraded.  It's possible there is a better way to do this, but essentially what I've done:

* `node.mode` is the `String` constant indicating what type of node this is, as we've always done
* `node.posix_mode` is the `Number` indicating the node type and permissions, as you'd expect from [`stat(2)`](http://man7.org/linux/man-pages/man2/stat.2.html)

On our `stats` object, I've exposed them as `type` (`node.mode`) and `mode` (`node.posix_mode`).  In reality, we should probably not expose `type`/`node.mode` at all, since the `Stats` object has functions to test node type in a platform independent way.  Unfortunately, our tests rely on this.  I'll file a bug to change to use the built-in getters instead.

@modeswitch see what you think of this.

### DIFF
--- a/src/directory-entry.js
+++ b/src/directory-entry.js
@@ -2,5 +2,5 @@ var NODE_TYPE_FILE = require('./constants.js').NODE_TYPE_FILE;
 
 module.exports = function DirectoryEntry(id, type) {
   this.id = id;
-  this.type = type || NODE_TYPE_FILE;
+  this.mode = type || NODE_TYPE_FILE;
 };
--- a/src/filesystem/implementation.js
+++ b/src/filesystem/implementation.js
@@ -110,7 +110,7 @@ function make_node(context, path, type, callback) {
   function create_node_in_parent(error, parentDirectoryNode) {
     if(error) {
       callback(error);
-    } else if(parentDirectoryNode.type !== NODE_TYPE_DIRECTORY) {
+    } else if(parentDirectoryNode.mode !== NODE_TYPE_DIRECTORY) {
       callback(new Errors.ENOTDIR('a component of the path prefix is not a directory', path));
     } else {
       parentNode = parentDirectoryNode;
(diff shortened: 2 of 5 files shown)


src/directory-entry.js
```diff
@@ -2,5 +2,5 @@ var NODE_TYPE_FILE = require('./constants.js').NODE_TYPE_FILE;
 
 module.exports = function DirectoryEntry(id, type) {
   this.id = id;
-  this.type = type || NODE_TYPE_FILE;
+  this.mode = type || NODE_TYPE_FILE;
```
modeswitch: This was previously `mode`, but changed to `type` in an earlier PR, is that right?

That change was not bundled into a release though?

humphd: That is correct.  I did it during the summer when I was writing the Plan 9 stuff with Filer.  We never shipped it to `npm` or offered built versions, so it's fine to revert it now and we won't affect anyone.

modeswitch: OK, that's what I thought.

modeswitch:
r+ from me. It looks like there's a conflict?

humphd:
Closing this in favour of https://github.com/filerjs/filer/pull/614, which is a lot simpler and less confusing to read.

PR #583, opened 2018-11-29
Improve coverage: add test for rel paths with symlink

humphd:
Looking through coverage data, I noticed we don't have a test for passing a relative path as `src` for a symlink.  I added support for this in the summer, but apparently didn't add a test.

I've also filled out out the docs to mention that we support it.


humphd:
That helped.  Landing.

PR #584, opened 2018-11-30
Increase coverage: check for non-int with chown/fchown

humphd:
We're missing tests for the case that `uid` and `gid` aren't integers with `chown` and `fchown`.  This improves coverage for those cases.


humphd:
Coverage increased, landing.

PR #587, opened 2018-12-01
Add Headless Firefox test support, fix failing truncate test in Firefox

humphd:
This adds support for running our tests in Firefox (headless) to match what we already do with Chrome (headless).  It's already found 1 bug! which I've fixed in this PR as well.  I've also added `npm` scripts for running the tests in each separate browser if people need to.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,16 @@
-sudo: true
-
 language: node_js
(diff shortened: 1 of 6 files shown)


humphd:
Improved coverage too! Merged.

Issue #561, opened 2018-11-15
Extend fs.access to support R_OK, W_OK, and X_OK


humphd:
Follow-up to #560 where we're adding initial support for `fs.access` with `F_OK`.  We'll need to extend this to also support checks for read, write and execute permissions.

deepanjali19:
Working on it! 👍

humphd:
Filed #601 for follow up.

PR #601, opened 2018-12-04
Fix for issue #561 - Extend fs.access to support R_OK, W_OK, and X_OK

deepanjali19:
I have added the support test for `fs.access()` to test various conditions with `R_OK`, `W_OK` and `X_OK`.

There is one test that is skipped for now. The test `should return an error if file is not executable and mode = X_OK`.

We create a new bug for this issue. Thank you 👍

### DIFF
--- a/src/filesystem/implementation.js
+++ b/src/filesystem/implementation.js
@@ -441,15 +441,17 @@ function make_directory(context, path, callback) {
 
 function access_file(context, path, mode, callback) {
   path = normalize(path);
-  find_node(context, path, function (err) {
+  find_node(context, path, function (err, node) {
     if (err) {
       return callback(err);
     }
-    /* 
-    TODO: we currently only support Constants.fsConstants.F_OK
-    Working to fix the issue: https://github.com/filerjs/filer/issues/561
-    */
-    callback(null);
+    if(mode === Constants.F_OK){
+      return callback(null);
+    }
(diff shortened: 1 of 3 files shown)


tests/spec/fs.access.spec.js
```diff
@@ -34,4 +34,96 @@ describe('fs.access', function () {
...
+      if (error) throw error;
+
+      fs.access('/myfile', fs.constants.W_OK, function (error) {
+        expect(error).not.to.exist;
+        done();
+      });
+    });
+  });
+
+  it.skip('should return an error if file is not executable and mode = X_OK', function (done) {
```
humphd: Can you add a comment above this that says:

```
// See bug https://github.com/filerjs/filer/issues/602
```

humphd: I filed https://github.com/filerjs/filer/issues/602 to deal with this later.

src/filesystem/implementation.js
```diff
@@ -441,15 +441,17 @@ function make_directory(context, path, callback) {
...
     }
-    /* 
-    TODO: we currently only support Constants.fsConstants.F_OK
-    Working to fix the issue: https://github.com/filerjs/filer/issues/561
-    */
-    callback(null);
+    if(mode === Constants.F_OK){
+      return callback(null);
+    }
+    if (!(mode & Constants.X_OK) || (node.mode & (Constants.fsConstants.S_IXUSR | Constants.fsConstants.S_IXGRP | Constants.fsConstants.S_IXOTH)))
```
humphd: Please wrap the `return` in a block:

```
if( ... ) {
  ...
}
```

humphd:
Great job @deepanjali19, you did it.

Issue #588, opened 2018-12-01
Remove fs.exists() console spam


humphd:
```
function exists(fs, context, path, callback) {
  function cb(err) {
    callback(err ? false : true);
  }
  console.warn('This method is deprecated. For more details see https://nodejs.org/api/fs.html#fs_fs_exists_path_callback');// eslint-disable-line no-console
  stat(fs, context, path, cb);
}
```

Let's get rid of the `console.warn()` line, it's just spamming the console.

yuzhouChen:
I will work on this

Issue #585, opened 2018-11-30
Remove duplicate stat.h constants defined in constants.js


humphd:
We are defining these twice (we can just use the ones node has on fs.constants);

```
  S_IFREG: 0x8000,
  S_IFDIR: 0x4000,
  S_IFLNK: 0xA000,
```

from https://github.com/filerjs/filer/blob/master/src/constants.js#L27-L29

and instead we can use them from https://github.com/filerjs/filer/blob/master/src/constants.js#L96

yuzhouChen:
I will work on this

PR #609, opened 2018-12-04
Fix issue-585: Remove duplicate stat.h constants defined in constants.js

yuzhouChen:
Fix [issue-585](https://github.com/filerjs/filer/issues/585)
After removing
```
  S_IFREG: 0x8000,
  S_IFDIR: 0x4000,
  S_IFLNK: 0xA000,
```
all tests are passed. Because I fixed the [issue-588](https://github.com/filerjs/filer/issues/588) and [issue-585](https://github.com/filerjs/filer/issues/585) almost at the same time, I found out that when I have the code without the [fix issue-588](https://github.com/filerjs/filer/pull/608), when I run ``npm test`` command, it cannot launched Firefox, as show in image below:

screen shot 2018-12-04 at 6 05 05 pm

So I think maybe have a relation between it. Just let you know that.


humphd:
You need to run `npm install` and also have Firefox installed to run Firefox headless.

yuzhouChen:
Fixed the callers in src/node.js to the correct version, let me know if I have done something wrong.

PR #610, opened 2018-12-05
Fix #603: add import/export provider to work with JSON fs images

### LINKED ISSUE #603: Create new memoy-based provider that can import/export JSON
Spin-off from https://github.com/filerjs/filer/issues/566#issuecomment-444218044: 

We need to subclass the `memory` provider and add new code that allows us to import/export JSON (encode binary data in base64).  The simplest thing to do here is to simply inherit from `memory` and make the following changes:

1. `constructor(name, jsonImage)` takes a `name` (like all providers), and a JSON filesystem image (`String`).
1. `.export()` returns a JSON filesystem image (`String`), where all binary data is encoded in base64.

We also need tests added for this provider.  Since we're mainly wanting this for testing purposes, the provider can live in the `tests/` directory somewhere.

humphd:
Fixes #603, sub-classing the `memory` provider to add import/export features for "binary" JSON (includes `Buffer`s).  Includes tests.

### DIFF
--- a/tests/lib/import-export-provider.js
+++ b/tests/lib/import-export-provider.js
@@ -1,0 +1,36 @@
+const MemoryProvider = require('../../src/providers/memory');
+const { parseBJSON } = require('../lib/test-utils');
+
+class ImportExportProvider extends MemoryProvider {
+
+  constructor(name, jsonImage) {
+    super(name);
+    this.unparsedJSONImage = jsonImage;
+  }
+
+  /**
+   * In addition to the usual setup of a Memory provider,
+   * also parse and overwrite the internal database.
+   */
+  open(callback) {
+    super.open(err => {
+      if(err) {
+        return callback(err);
+      }
+
+      try {
+        this.db = parseBJSON(this.unparsedJSONImage);
+        this.unparsedJSONImage = null;
+        callback();
+      } catch(e) {
+        callback(new Error(`unable to parse JSON filesystem image: ${e.message}`));
+      }
+    });
+  }
+
+  export() {
+    return JSON.stringify(this.db);
+  }
+}
+
+module.exports = ImportExportProvider;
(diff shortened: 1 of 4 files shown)


tests/lib/import-export-provider.js
```diff
@@ -0,0 +1,36 @@
+const MemoryProvider = require('../../src/providers/memory');
+const { parseBJSON } = require('../lib/test-utils');
```
modeswitch: Nice, good choice :)

modeswitch: You may want to consider [this](https://www.npmjs.com/package/bson).

tests/lib/import-export-provider.js
```diff
@@ -0,0 +1,36 @@
...
+
+class ImportExportProvider extends MemoryProvider {
```
modeswitch: Can you call this `SerializableMemoryProvider` instead?

modeswitch:
Feel free to keep merging stuff. I don't want to hold you up. I'll leave comments and you can address them in follow-up patches if you like.

humphd:
Perfect, leave comments and I'll fix in follow-ups.

I'll do another PR now to fix what you suggest above.

tests/lib/import-export-provider.js
modeswitch: Never mind, I didn't realize that `JSON.stringify` would serialize `Buffer`s!

humphd:
I tried to use `js-bson` instead, but I can't figure out how to get back to a pure JS object from the decorated type:

```
> var BSON = require('bson')
> var o = { b: Buffer.from('data') }
> o
{ b: <Buffer 64 61 74 61> }
> BSON.serialize(o)
<Buffer 11 00 00 00 05 62 00 04 00 00 00 00 64 61 74 61 00>
> var bson = BSON.serialize(o)
> p = BSON.deserialize(bson)
{ b: Binary { sub_type: 0, position: 4, buffer: <Buffer 64 61 74 61> } }
```

How would I get `o` back from this BSON document?  Using `deserialize(o)` gives me a decorated type.

modeswitch:
I think what you're doing is fine. I didn't know that `JSON.stringify` would handle `Buffer` objects in Node. And your handler for `JSON.parse` should be fine too.

humphd:
Nit fixed in https://github.com/filerjs/filer/pull/611.

PR #613, opened 2018-12-06
Fix #604: cli for buidling JSON image from folder

### LINKED ISSUE #604: Create a tool in node that create an importable JSON disk image
Spin-off from https://github.com/filerjs/filer/issues/566#issuecomment-444218044, and building on #603:

We need a node cli that takes a path to a folder on disk, creates a `Filer` instance with the `memory-import-export` provider (see #603), bridges node's `fs` module to read all the files beneath the folder into the `Filer` instance, then `exports()` the contents of the provider's backing to JSON to a file on disk.

```
$ node create-image filesystem/ ./filesystem.json
```

This will take the contents of the folder `./filesystem` and read all directories, files, symlinks, etc., then produce a JSON file suitable for being imported into the provider from #603.

humphd:
Adds a new node cli tool, `tools/fs-image`, that can take a directory, and use it to build a Filer filesystem image as a JSON file.  If an export filename is given as the second arg, it will be used; otherwise a name will be generated.

```
$ node tools/fs-image.js ./lib --verbose
Writing file lib/async.js to /async.js [2.4 kB]
  File Node: mode=16877 uid=501 gid=20 atime=1544073922000 mtime=1543523556000
Writing file lib/eventemitter.js to /eventemitter.js [1.99 kB]
  File Node: mode=16877 uid=501 gid=20 atime=1544073922000 mtime=1543523556000
Writing file lib/intercom.js to /intercom.js [7.32 kB]
  File Node: mode=16877 uid=501 gid=20 atime=1544073922000 mtime=1543523556000
Wrote filesystem JSON to filesystem.json
```

The file `filesystem.json` is ready to be loaded into the `SerializableMemoryProvider` and used.

NOTE: I had to make one adjustment to the `SerializableMemoryProvider` to allow it to start without an image (e.g., case that you want to create vs. consume an image).

I haven't tested this with symlinks or hardlinks.  We can do that in follow-ups as we need them.


humphd:
I need to add some more to this before I land it.  I'll iterate on this a bit.

PR #632, opened 2018-12-18
Fix #605: add migration tests for Filer 0.43

### LINKED ISSUE #605: Allow test harness to load a JSON filesystem image from disk into browser
Spin-off from https://github.com/filerjs/filer/issues/566#issuecomment-444218044, and building on #603, #604:

We need to teach our test harness how to load a JSON filesystem image from disk into the browser, and get it loaded into the provider being created in #603.  We should be able to use this to set the initial state of a Filer filesystem instance before a set of tests runs.

A simple test for this might include creating a tiny filesystem JSON file with a single file, and having a test that confirms that the file exists and the metadata/data for the file is what we expect via Filer.

humphd:
This adds initial support for migration testing, fixing #605.  The basic idea is that we be able to test whether a filesystem that was created under Filer 0.43 (current release on npm) could be mounted and used from the current version.  Here are some important things to note:

1. All of the files related to what I'm doing live under `tests/filesystems/*`
1. I've created a tiny filesystem layout in `tests/filesystem/tiny-fs` with a few files and folders.
1. I've created a JSON filesystem image in `tests/filesystems/images/tiny-fs.0.43.json` based on `tests/filesystem/tiny-fs` and using [this version](https://github.com/filerjs/filer/blob/d66114e20c7f0235698d9933dbd90217ba86fa4e/dist/filer.min.js) of Filer.  Here's how you do that:
    ```node tools/fs-image.js -f tools/filer-0.43.min.js -v tests/filesystems/tiny-fs```
1. I've created our first migration test in `tests/filesystems/migrations/from-0.43.test.js`, which tries to open the filesystem image `tests/filesystems/images/tiny-fs.0.43.json` and use it (read files, examine dirs).  We could add more tasks/checks to this, but this is a start.

The good news is, this has already confirmed my suspicions, and `master` can't mount a filesystem created with `0.43`.  My PR #614 should fix that, but I'll land this first, and then I can know for sure!


humphd:
I forgot to mention, I also added a `npm run test:migrations` script.  Since it will fail at the moment, I haven't wired it into our default `test` run yet; but I will once they pass.

PR #640, opened 2018-12-20
Automate releases to git, GitHub, npm

### LINKED ISSUE #404: Automate rebuilds of dist/ when we land code on master
We keep a pre-built version of Filer's bundle in `dist/`.  Currently, it will lag behind the code on `master` unless manually updated.  This is error prone, since people will forget, or not know to do this.  We should consider letting Travis do it for us when we're merging onto `master`.

Similarly, once we have that, we should also tackle `npm` republishing of the module.

humphd:
Fixes #404 by modernizing our release process to use [release-it](https://www.npmjs.com/package/release-it).  With the changes in this PR, one of @modeswitch or me (or whoever we give rights to), can do:

```bash
$ npm run release
```

This will do a bunch of things automatically:

1. run the tests
1. build `dist/`
1. bump the version in `package.json`
1. tag the commit
1. commit everything
1. push the commit and tag
1. create a GitHub Release, with changelog
1. push a new package to npm

See [this post](https://medium.com/@webprolific/using-release-it-60b96515c073) for a more verbose description. The tool is interactive by default, and will look something like this:

All default choices are defined in `.release-it.json` below.

I've included docs in `CONTRIBUTING.md`, but to summarize what is important:

1. each dev doing releases will need to copy `env.sample` to `.env` and include a `GITHUB_TOKEN` they [generate for themselves](https://help.github.com/articles/creating-a-personal-access-token-for-the-command-line/).
1. `npm login` so they can publish to `npm`

I haven't run this fully yet, so I may need to fix a few corner cases.  We can do a few runs to get it right, then start using this.


humphd:
You can see an example of what the GitHub Releases end up looking like here https://github.com/webpro/release-it/releases.

@modeswitch my thinking is to land this, and try doing a release, tweak things, repeat until I get it right.  If you disagree, speak up!

My personal view is that I can probably just do minor updates, since I've been able to not break 0.44 versions.  However, tons has changed, so I could also do a major number update too.  Let me know your thoughts.

humphd:
Merged.

PR #651, opened 2019-01-03
Add Filer.fs to support creating a filesystem like node.js

humphd:
To be more in sync with how node.js works, this adds `Filer.fs`, which lazily creates a filesystem instance.  It also adds `Filer.path` to match what node devs would expect, and allows users to do the following:

```js
const { fs, Buffer, path } = require('filer');
```

I have kept `Filer.Path` to not break the API, but we could remove it later if we wanted.

I also spotted a stray WebSQL reference in the tests, which this also removes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,20 +81,28 @@ backend storage providers, for example `Memory`. See the section on [Storage Pro
 <a name="overviewExample"></a>
 
 ```js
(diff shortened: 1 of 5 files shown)


humphd:
NOTE: the lower coverage on `src/index.js` is because we can't run this in our current coverage environment (node) because it depends on indexeddb--it's fine.

PR #656, opened 2019-01-03
Add test API to shim indexedDB when necessary for better coverage

humphd:
Shim IndexedDB in tests when we have to in order to get better test coverage in node.js (where it doesn't exist).


humphd:
To make this work the way I wanted, I had to alter how we test for, and access, `indexedDB` on the global to use late binding vs. assuming it exists at startup.

PR #668, opened 2019-01-24
Fixes #662: exchange var for const of require statements in src/filesystem/interface.js

### LINKED ISSUE #662: Update code in src/filesystem/interface.js to use const rather than var
Currently interface.js is using many var statements to import files such as in line 1: 
`var { promisify } = require('es6-promisify');`

these should instead be included using const

trowacat:
Changed decleration of import files from var to const in interface.js

### DIFF
--- a/src/filesystem/interface.js
+++ b/src/filesystem/interface.js
@@ -1,15 +1,9 @@
+'use strict';
+
 const { promisify } = require('es6-promisify');
 
 const Path = require('../path.js');
 const nop = require('../shared.js').nop;
-
-const Constants = require('../constants.js');
-const FILE_SYSTEM_NAME = Constants.FILE_SYSTEM_NAME;
-const FS_FORMAT = Constants.FS_FORMAT;
-const FS_READY = Constants.FS_READY;
-const FS_PENDING = Constants.FS_PENDING;
-const FS_ERROR = Constants.FS_ERROR;
-const FS_NODUPEIDCHECK = Constants.FS_NODUPEIDCHECK;
 
 const providers = require('../providers/index.js');
 
@@ -19,20 +13,28 @@ const FSWatcher = require('../fs-watcher.js');
 const Errors = require('../errors.js');
 const defaultGuidFn = require('../shared.js').guid;
 
-const STDIN = Constants.STDIN;
-const STDOUT = Constants.STDOUT;
-const STDERR = Constants.STDERR;
+const {
+  FILE_SYSTEM_NAME,
+  FS_FORMAT,
+  FS_READY,
+  FS_PENDING,
+  FS_ERROR,
+  FS_NODUPEIDCHECK,
+  STDIN,
+  STDOUT,
+  STDERR
+} = require('../constants.js');
 
 // The core fs operations live on impl
 const impl = require('./implementation.js');
 
 // node.js supports a calling pattern that leaves off a callback.
 function maybeCallback(callback) {
-  if(typeof callback === 'function') {
+  if (typeof callback === 'function') {
     return callback;
   }
-  return function(err) {
-    if(err) {
+  return function (err) {
+    if (err) {
       throw err;
     }
   };
@@ -40,7 +42,7 @@ function maybeCallback(callback) {
 
 // Default callback that logs an error if passed in
 function defaultCallback(err) {
-  if(err) {
+  if (err) {
     /* eslint no-console: 0 */
     console.error('Filer error: ', err);
   }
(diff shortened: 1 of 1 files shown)


andrewkoung:
I don't disagree with any of the changes made.

src/filesystem/interface.js
```diff
@@ -1,30 +1,30 @@
...
+const nop = require('../shared.js').nop;
 
-var Constants = require('../constants.js');
-var FILE_SYSTEM_NAME = Constants.FILE_SYSTEM_NAME;
-var FS_FORMAT = Constants.FS_FORMAT;
-var FS_READY = Constants.FS_READY;
-var FS_PENDING = Constants.FS_PENDING;
-var FS_ERROR = Constants.FS_ERROR;
-var FS_NODUPEIDCHECK = Constants.FS_NODUPEIDCHECK;
+const Constants = require('../constants.js');
```
humphd: Let's use destructing assignment here, since we're pulling a bunch of stuff off `Constants`:

```js
const {
  FILE_SYSTEM_NAME,
  FS_FORMAT,
  ...
} = require('../constants.js');
```

src/filesystem/interface.js
```diff
@@ -1,30 +1,30 @@
-var { promisify } = require('es6-promisify');
```
humphd: Can you add `use strict`; to the top of this file please?

trowacat:
@humphd Thank you for the recommendations,  I've made all the additions you requested. Let me know if there is still more that needs to be done!

src/filesystem/interface.js
```diff
@@ -1,67 +1,70 @@
...
-
-var STDIN = Constants.STDIN;
-var STDOUT = Constants.STDOUT;
-var STDERR = Constants.STDERR;
+'use strict';
+
+const { promisify } = require('es6-promisify');
+
+const Path = require('../path.js');
+const nop = require('../shared.js').nop;
```
humphd: Since we pull two things out of `shared.js` (here and below on line 15), let's do ti all together:

```js
const {
  nop,
  guid: defaultGuidFn
} = require('../shared.js');

File: src/filesystem/interface.js [hunk 4]
```diff
@@ -1,67 +1,70 @@
...
+  FILE_SYSTEM_NAME,
+  FS_FORMAT,
+  FS_READY,
+  FS_PENDING,
+  FS_ERROR,
+  FS_NODUPEIDCHECK,
+  STDIN,
+  STDOUT,
+  STDERR
+} = Constants;
```
humphd: Merge line 10 into this, and just do:

```js
} = require('../constants.js');
```

src/filesystem/interface.js
```diff
@@ -1,67 +1,70 @@
...
 
 // The core fs operations live on impl
-var impl = require('./implementation.js');
+const impl = require('./implementation.js');
 
 // node.js supports a calling pattern that leaves off a callback.
 function maybeCallback(callback) {
-  if(typeof callback === 'function') {
+  if (typeof callback === 'function') {
```
humphd: In general, be careful of this kind of thing (i.e., changing formatting in lines unrelated to your actual changes).  We call this a "spurious whitespace change" and open source projects tend to want to avoid it, since it masks the history of this line of code, making it look like it was changed here by you, when actually it was just a formatting fix.

I'll allow it this time, but be aware that it's not generally a good idea.  If your editor did it automatically, I'd disable that.

src/filesystem/interface.js
```diff
@@ -133,14 +136,14 @@ function FileSystem(options, callback) {
...
+  const guid = options.guid ? options.guid : defaultGuidFn;
+  const provider = options.provider || new providers.Default(options.name || FILE_SYSTEM_NAME);
   // If we're given a provider, match its name unless we get an explicit name
-  var name = options.name || provider.name;
-  var forceFormatting = flags.includes(FS_FORMAT);
+  const name = options.name || provider.name;
+  const forceFormatting = flags.includes(FS_FORMAT);
 
-  var fs = this;
+  let fs = this;
```
humphd: This can likely be `const`.  The `fs` instance won't change during this scope, will it?

src/filesystem/interface.js
```diff
@@ -285,18 +288,18 @@ function FileSystem(options, callback) {
       callback(error, fs);
     }
 
-    if(err) {
+    if (err) {
       return complete(err);
     }
 
-    var context = provider.getReadWriteContext();
+    let context = provider.getReadWriteContext();
```
humphd: This won't change either, you can use `const` I think.

src/filesystem/interface.js
```diff
@@ -334,64 +337,64 @@ function FileSystem(options, callback) {
...
+    const methodName = method.name;
+    const shouldPromisify = method.promises === true;
 
-    FileSystem.prototype[methodName] = function() {
-      var fs = this;
-      var args = Array.prototype.slice.call(arguments, 0);
-      var lastArgIndex = args.length - 1;
+    FileSystem.prototype[methodName] = function () {
+      const fs = this;
+      let args = Array.prototype.slice.call(arguments, 0);
```
humphd: `const`

trowacat:
@humphd Thank you again for the input! I believe all is in order now.

PR #669, opened 2019-01-24
Fixes #661: replaced var with const

### LINKED ISSUE #661: Modernizing (tests/spec/fs.symlink.spec.js)
Modern JavaScript prefers that programmers use let, const, and strict mode instead of var.

andrewkoung:
Changed the every keyword "var" in the whole file to use const.


andrewkoung:
@humphd changes made

jatinkumarg:
Great, I have also added ```'use strict'``` at the top of the fs.readdir.spec.js file.

PR #676, opened 2019-01-25
Fixes Issue#674: Refactored var declarations to const or let declarations to address i…

ndngo:
…issue#674

Added the 'use strict' declaration at the top and refactored var variable declarations to use const or let.


humphd:
Agree that `const` is best when the variable won't change values (e.g., `fs` will always be the same thing within each test scope).  Save `let` for cases where it will change.

ndngo:
That makes sense. I have changed variables that don't change values to be declared with const.

Issue #667, opened 2019-01-24
Update "filer/tests/spec/fs.readdir.spec.js" to use let/const instead of var


jatinkumarg:
I just noticed that "fs.readdir.spec.js" contains lots of var declared variables. Can I get the authorization to update the required variables to use const or let instead of using var?

humphd:
Go ahead @jatinkumarg, it's all yours.

PR #678, opened 2019-01-26
fixes issue #667

jadach1:
Update code in tests/some/fs.write.spec.js to use const/let AND strict mode

### DIFF
--- a/tests/spec/fs.write.spec.js
+++ b/tests/spec/fs.write.spec.js
@@ -1,24 +1,26 @@
-var util = require('../lib/test-utils.js');
-var expect = require('chai').expect;
+'use strict';
+
+let util = require('../lib/test-utils.js');
+let expect = require('chai').expect;
 
 describe('fs.write', function() {
   beforeEach(util.setup);
   afterEach(util.cleanup);
 
   it('should be a function', function() {
-    var fs = util.fs();
+    let fs = util.fs();
     expect(fs.write).to.be.a('function');
   });
 
   it('should error if file path is undefined', function() {
-    var fs = util.fs();
-    var fn = () => fs.writeFile(undefined, 'data');
+    let fs = util.fs();
+    let fn = () => fs.writeFile(undefined, 'data');
     expect(fn).to.throw();
   });
 
   it('should write data to a file', function(done) {
-    var fs = util.fs();
-    var buffer = Buffer.from([1, 2, 3, 4, 5, 6, 7, 8]);
+    let fs = util.fs();
+    let buffer = Buffer.from([1, 2, 3, 4, 5, 6, 7, 8]);
 
     fs.open('/myfile', 'w', function(error, fd) {
       if(error) throw error;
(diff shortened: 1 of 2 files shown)


tests/spec/fs.write.spec.js
```diff
@@ -1,24 +1,25 @@
...
+const util = require('../lib/test-utils.js');
+const expect = require('chai').expect;
 
 describe('fs.write', function() {
   beforeEach(util.setup);
   afterEach(util.cleanup);
 
   it('should be a function', function() {
-    var fs = util.fs();
+    let fs = util.fs();
```
vlogozzo: fs can be const as it isnt changed in this scope
```suggestion
    const fs = util.fs();
```

tests/spec/fs.write.spec.js
```diff
@@ -1,24 +1,25 @@
...
     expect(fs.write).to.be.a('function');
   });
 
   it('should error if file path is undefined', function() {
-    var fs = util.fs();
-    var fn = () => fs.writeFile(undefined, 'data');
+    let fs = util.fs();
+    let fn = () => fs.writeFile(undefined, 'data');
```
vlogozzo: fn can be const as well
```suggestion
    const fn = () => fs.writeFile(undefined, 'data');
```

tests/spec/fs.write.spec.js
```diff
@@ -1,24 +1,25 @@
...
     expect(fn).to.throw();
   });
 
   it('should write data to a file', function(done) {
-    var fs = util.fs();
-    var buffer = Buffer.from([1, 2, 3, 4, 5, 6, 7, 8]);
+    let fs = util.fs();
+    let buffer = Buffer.from([1, 2, 3, 4, 5, 6, 7, 8]);
```
vlogozzo: buffer is just passed into a function it can be const
```suggestion
    constt buffer = Buffer.from([1, 2, 3, 4, 5, 6, 7, 8]);
```

tests/spec/fs.write.spec.js
```diff
@@ -1,24 +1,25 @@
...
   it('should be a function', function() {
-    var fs = util.fs();
+    const fs = util.fs();
     expect(fs.write).to.be.a('function');
   });
 
   it('should error if file path is undefined', function() {
-    var fs = util.fs();
-    var fn = () => fs.writeFile(undefined, 'data');
+    let fs = util.fs();
```
humphd: `const` here, since `fs` won't change

tests/spec/fs.write.spec.js
```diff
@@ -1,24 +1,25 @@
...
+    const fn = () => fs.writeFile(undefined, 'data');
     expect(fn).to.throw();
   });
 
   it('should write data to a file', function(done) {
-    var fs = util.fs();
-    var buffer = Buffer.from([1, 2, 3, 4, 5, 6, 7, 8]);
+    let fs = util.fs();
```
humphd: `const` here too

tests/spec/fs.write.spec.js
```diff
@@ -38,9 +39,9 @@ describe('fs.write', function() {
   });
 
   it('should update the current file position', function(done) {
-    var fs = util.fs();
-    var buffer = Buffer.from([1, 2, 3, 4, 5, 6, 7, 8]);
-    var _result = 0;
+    let fs = util.fs();
```
humphd: And here, `const`

PR #680, opened 2019-01-27
Changed fs.watch.spec.js according to filed issue-666

vrozin:
The file had its variable being declared with usage of "var" -> changed to "let" and "const" where applicable; added "use strict" at global scope.

### DIFF
--- a/tests/spec/fs.watch.spec.js
+++ b/tests/spec/fs.watch.spec.js
@@ -1,5 +1,7 @@
-var util = require('../lib/test-utils.js');
-var expect = require('chai').expect;
+'use strict';
+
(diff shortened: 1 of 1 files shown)


tests/spec/fs.watch.spec.js
```diff
@@ -126,4 +128,4 @@ describe('fs.watch', function() {
       });
     });
   });
-});
+});
```
humphd: One tip: it's always a good idea to end files with a newline.  We do this for cases where files get processed/concatenated together, and you want to avoid the case that the last line of one file, and the first line of the next, get joined and cause an error.

Issue #670, opened 2019-01-25
Update code in filer/src/stats.js to use const instead of var in loading modules


AlvinValdez:
var Constants = require('./constants.js');
var Path = require('./path.js');

Should replace the var with const

humphd:
@AlvinValdez sure, you can work on this.  NOTE: when you file an issue, you need a description to go with it, vs. just a title.  Can you edit this issue (three dots on the top-right of your original comment, and then Edit) to add one?

PR #687, opened 2019-01-28
Fix for issue-670: Changed var to const

AlvinValdez:
Updated code in filer/src/stats.js to use const instead of var.


yuansheng1989:
You can add 'use strict' in the top of the file

humphd:
One tip: instead of `Fix for issue-670` use `Fixes #670` and GitHub will autoclose the issue when this gets merged.  See https://help.github.com/articles/closing-issues-using-keywords/

PR #688, opened 2019-01-28
Fixed #665: Replaced var with let or const and added strict mode fs.a…

### LINKED ISSUE #665: Update the code inside filer/tests/spec/fs.access.spec.js to use let and const, and strict mode.
1. Replace let and const with var inside the code
2. Use strict mode

ithompson4:
Updated the code in filer/tests/spec/fs.access.spec.js to fix issue #665:
- replaced var with let and const
- added strict mode

### DIFF
--- a/tests/spec/fs.access.spec.js
+++ b/tests/spec/fs.access.spec.js
@@ -8,7 +8,7 @@ describe('fs.access', function () {
   afterEach(util.cleanup);
 
   it('should expose access mode flags on fs and fs.constants', function() {
-    let fs = util.fs();
+    const fs = util.fs();
 
     // F_OK
     expect(fs.F_OK).to.equal(0);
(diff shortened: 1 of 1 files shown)


tests/spec/fs.access.spec.js
```diff
@@ -26,12 +28,12 @@ describe('fs.access', function () {
   });
 
   it('should be a function', function () {
-    var fs = util.fs();
+    let fs = util.fs();
```
yuansheng1989: use const is better, because it's not changed

tests/spec/fs.access.spec.js
```diff
@@ -26,12 +28,12 @@ describe('fs.access', function () {
...
     expect(typeof fs.access).to.equal('function');
   });
 
   it('should return an error if file does not exist', function (done) {
-    var fs = util.fs();
+    let fs = util.fs();
```
yuansheng1989: use const is better, because it's not changed

tests/spec/fs.access.spec.js
```diff
@@ -41,8 +43,8 @@ describe('fs.access', function () {
   });
 
   it('should return no error if file does exist and mode = F_OK', function (done) {
-    var fs = util.fs();
-    var contents = 'This is a file.';
+    let fs = util.fs();
```
humphd: You can follow the same pattern as above, and use `const` here, since `fs` won't change during the scope of this function.

ithompson4: I replaced all let statements with const and pushed another commit.

PR #693, opened 2019-01-28
Update fs.spec.js to use strict mode and changed var to let

OnesQuared:
Updated to use strict mode to detect silent errors
changed the var to let in order to be transferable in the future

### DIFF
--- a/tests/spec/fs.spec.js
+++ b/tests/spec/fs.spec.js
@@ -9,13 +9,13 @@ describe('fs', function() {
   afterEach(util.cleanup);
 
   it('is an object', function() {
-    let fs = util.fs();
+    const fs = util.fs();
     expect(typeof fs).to.equal('object');
     expect(fs).to.be.an.instanceof(Filer.FileSystem);
   });
 
   it('should have a root directory', function(done) {
-    let fs = util.fs();
+    const fs = util.fs();
     fs.stat('/', function(error, result) {
       expect(error).not.to.exist;
       expect(result).to.exist;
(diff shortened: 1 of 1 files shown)


woosle1234:
For the require lines at the top of the file it's better to use const instead of let especially since those are global declarations

tests/spec/fs.spec.js
```diff
@@ -1,19 +1,21 @@
-var Filer = require('../../src');
-var util = require('../lib/test-utils.js');
-var expect = require('chai').expect;
```
woosle1234: Should change to const

OnesQuared:
@humphd @woosle1234 The global variables have been changed as per request, please review the code again and let me know if anything else would need to be changed

tests/spec/fs.spec.js
```diff
@@ -1,19 +1,21 @@
...
+const util = require('../lib/test-utils.js');
+const expect = require('chai').expect;
 
 describe('fs', function() {
   beforeEach(util.setup);
   afterEach(util.cleanup);
 
   it('is an object', function() {
-    var fs = util.fs();
+    let fs = util.fs();
```
humphd: `const` here too, `fs` never changes.

tests/spec/fs.spec.js
```diff
@@ -1,19 +1,21 @@
...
     expect(typeof fs).to.equal('object');
     expect(fs).to.be.an.instanceof(Filer.FileSystem);
   });
 
   it('should have a root directory', function(done) {
-    var fs = util.fs();
+    let fs = util.fs();
```
humphd: And here, `const`, `fs` never changes.

Issue #660, opened 2019-01-23
update tests/spec/fs.readlink.spec.js to use const, let and strict mode       


yuansheng1989:
fs.readlink.spec.js is using var. It should be replaced by let or const in strict mode.

humphd:
Sounds good @yuansheng1989, it's all yours.

humphd:
NOTE: when you do a PR, you can tell GitHub to automatically close the associated issue by using "Fixes #660" in your PR description.

PR #694, opened 2019-01-29
Fix issue #660: Updated the code to replace var with let, const in strict mode

yuansheng1989:
- "var" was replaced with "let" or"const"
-  "use strict" is added to the file

### DIFF
--- a/tests/spec/fs.readlink.spec.js
+++ b/tests/spec/fs.readlink.spec.js
@@ -1,17 +1,19 @@
-var util = require('../lib/test-utils.js');
-var expect = require('chai').expect;
+'use strict';
+
+const util = require('../lib/test-utils.js');
+const expect = require('chai').expect;
 
 describe('fs.readlink', function() {
   beforeEach(util.setup);
   afterEach(util.cleanup);
 
   it('should be a function', function() {
-    var fs = util.fs();
+    let fs = util.fs();
     expect(fs.readlink).to.be.a('function');
   });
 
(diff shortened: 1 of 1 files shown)


OlenaVyshnevska:
Hey, I think it might be better to use const instead of let for fs and contents variables cause we aren't changing their values

tests/spec/fs.readlink.spec.js
```diff
@@ -1,17 +1,19 @@
...
+const util = require('../lib/test-utils.js');
+const expect = require('chai').expect;
 
 describe('fs.readlink', function() {
   beforeEach(util.setup);
   afterEach(util.cleanup);
 
   it('should be a function', function() {
-    var fs = util.fs();
+    let fs = util.fs();
```
OlenaVyshnevska: You might want to change let to const as we aren't changing "fs" value

tests/spec/fs.readlink.spec.js
```diff
@@ -1,17 +1,19 @@
...
     expect(fs.readlink).to.be.a('function');
   });
 
   it('should return an error if part of the parent destination path does not exist', function(done) {
-    var fs = util.fs();
```
OlenaVyshnevska: Same here

tests/spec/fs.readlink.spec.js
```diff
@@ -1,17 +1,19 @@
...
+    let fs = util.fs();
```
OlenaVyshnevska: Same here

Issue #681, opened 2019-01-27
Update filer/tests/spec/filer.filesystem.spec.js to use let instead of var


woosle1234:
I will change the code to use let instead of var for non global variable declarations and changing require statements to const

humphd:
Sounds good @woosle1234.

woosle1234:
https://github.com/filerjs/filer/pull/695

PR #695, opened 2019-01-29
Fix issue #681: Updated filer/tests/spec/filer.filesystem.spec.js

woosle1234:
require lines use const instead of var
variable declaration in local scopes use let instead of var

### DIFF
--- a/tests/bugs/issue681.js
+++ b/tests/bugs/issue681.js
@@ -2,7 +2,7 @@ const Filer = require('../../src');
 const util = require('../lib/test-utils.js');
 const expect = require('chai').expect;
 
-describe('Filer.FileSystem', function() {
+describe('Filer.FileSystem - Issue 681', function() {
   beforeEach(util.setup);
   afterEach(util.cleanup);
 
--- a/tests/index.js
+++ b/tests/index.js
@@ -85,6 +85,7 @@ require('./bugs/issue267.js');
 require('./bugs/issue270.js');
 require('./bugs/rename-dir-trailing-slash.js');
 require('./bugs/issue357.js');
+require('./bugs/issue681.js');
 
 // Sample code from README
 require('./spec/readme.example.spec');


woosle1234:
> Hi, I think you are missing 'use strict' at the top of your file.

Added

tests/bugs/issue681.js
```diff
@@ -0,0 +1,46 @@
+const Filer = require('../../src');
```
SahibArora: You can also use 
``` JS
'use strict';
```
know more about strict here > https://www.geeksforgeeks.org/strict-mode-javascript/

tests/index.js
```diff
@@ -85,6 +85,7 @@ require('./bugs/issue267.js');
 require('./bugs/issue270.js');
 require('./bugs/rename-dir-trailing-slash.js');
 require('./bugs/issue357.js');
+require('./bugs/issue681.js');
```
humphd: You don't need to add a new file for this bug.  You're updating the code in `tests/spec/filer.filesystem.spec.js`, which already exists, and you've addressed below.  You can remove this and `tests/bugs/issue681.js`:

```
git rm tests/bugs/issue681.js
git checkout master tests/index.js
```

This will undo your changes.

tests/spec/filer.filesystem.spec.js
```diff
@@ -1,16 +1,17 @@
...
+const util = require('../lib/test-utils.js');
+const expect = require('chai').expect;
 
 describe('Filer.FileSystem', function() {
   beforeEach(util.setup);
   afterEach(util.cleanup);
 
   it('should properly mount new or existing filesystem', function(done) {
-    var provider = util.provider().provider;
+    let provider = util.provider().provider;
```
humphd: This can be `const` since `provider` will never change.

tests/spec/filer.filesystem.spec.js
```diff
@@ -1,16 +1,17 @@
...
 
     // 1) Should be able to open a new filesystem, and get empty root
-    var fs1 = new Filer.FileSystem({provider: provider}, function() {
+    let fs1 = new Filer.FileSystem({provider: provider}, function() {
```
humphd: Same thing here, use `const`

tests/spec/filer.filesystem.spec.js
```diff
@@ -20,7 +21,7 @@ describe('Filer.FileSystem', function() {
           if(err) throw err;
 
           // 2) Should be able to open an existing filesystem
-          var fs2 = new Filer.FileSystem({provider: provider}, function() {
+          let fs2 = new Filer.FileSystem({provider: provider}, function() {
```
humphd: `const`

tests/spec/filer.filesystem.spec.js
```diff
@@ -29,7 +30,7 @@ describe('Filer.FileSystem', function() {
 
 
               // 3) FORMAT flag should wipe an existing filesystem
-              var fs3 = new Filer.FileSystem({provider: provider, flags: ['FORMAT']}, function() {
+              let fs3 = new Filer.FileSystem({provider: provider, flags: ['FORMAT']}, function() {
```
humphd: `const`

woosle1234: changed

tests/spec/filer.filesystem.spec.js
woosle1234: changed

tests/spec/filer.filesystem.spec.js
woosle1234: changed

tests/spec/filer.filesystem.spec.js
woosle1234: changed

tests/index.js
woosle1234: Removed file and require line

woosle1234:
> This is coming along well. There are two main things we need to do to finish it.
> 
> 1. I think you've gotten confused about what's required for tests with a bug.  You've added `tests/bugs/issue681.js`, which isn't actually necessary in this case.  Sometimes we do this (i.e., add a test for a particular bug), because we are fixing a problem, and want to add a test to make sure it doesn't get added back again.  The tests in `tests/bugs/*` are all specific to old bugs we don't want to regress.  I've left some comments for how to undo this.
> 2. You need to make some changes to use `const` in the final file.  What you did is actually correct, but you can use `const` because those variables won't every change.

Changes implemented in commit "issue 681 revision v2"

Issue #692, opened 2019-01-28
Update code in filer/tests/spec/time-flags.spec.js to use "const"/"let" instead of "var"


Violet-XiaoWeiHuang:
I will update variable declarations to use "const"/"let" instead of of "var" based of the issue.

humphd:
This is merged now, closing.

PR #696, opened 2019-01-29
Fix ssue #692: Update code in filer/tests/spec/time-flags.spec.js

Violet-XiaoWeiHuang:
"var" was replaced by "let", as the passed variable may be modified later. In this case, I avoid using "const".

### DIFF
--- a/tests/spec/time-flags.spec.js
+++ b/tests/spec/time-flags.spec.js
@@ -1,14 +1,14 @@
-var Filer = require('../../src');
-var util = require('../lib/test-utils.js');
-var expect = require('chai').expect;
+let Filer = require('../../src');
+let util = require('../lib/test-utils.js');
+let expect = require('chai').expect;
 
 describe('node times (atime, mtime, ctime) with mount flags', function() {
 
-  var dirname = '/dir';
-  var filename = '/dir/file';
+  let dirname = '/dir';
+  let filename = '/dir/file';
 
   function memoryFS(flags, callback) {
-    var name = util.uniqueName();
+    const name = util.uniqueName();
     return new Filer.FileSystem({
       name: name,
       flags: flags || [],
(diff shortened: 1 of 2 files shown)


rscotchmer:
Hi!  It looks like you're off to a great start, but I think there are some changes to be made yet.  [On line 13](https://github.com/filerjs/filer/blob/31fd258722ffe63075649dc062414f60ead59171/tests/spec/time-flags.spec.js#L13), you've got a comma instead of a semi-colon, which I think is why your Travis CI build is failing -- I copied your code into my local version of filer, and was able to run the tests perfectly after I fixed that.

You're also missing 'use strict' at the top of your file, which you need to use so the file runs in strict mode.  I think you could also change the 3 "lets" on lines 3, 4, and 5 to "const"; the values there should never change, so it's safe to use const.

Violet-XiaoWeiHuang:
Thank you for your comment. It's strange I tried locally, all tests passed. 
And [in line 13](https://github.com/filerjs/filer/blob/31fd258722ffe63075649dc062414f60ead59171/tests/spec/time-flags.spec.js#L13) you mentions the comma should be correct as name is one of the data member.  @rscotchmer

rscotchmer:
I feel you -- I had all tests passing locally, and then when I put up the pull request it decided to fail a few -- the joys of programming, eh?  I'm glad all the tests are passing for you now, though!

You're definitely right about that comma!  My bad, I should've taken a closer look at that one.

Violet-XiaoWeiHuang:
Thank you @ithompson4.

tests/spec/time-flags.spec.js
```diff
@@ -1,14 +1,16 @@
...
+
+const Filer = require('../../src');
+const util = require('../lib/test-utils.js');
+const expect = require('chai').expect;
 
 describe('node times (atime, mtime, ctime) with mount flags', function() {
 
-  var dirname = '/dir';
-  var filename = '/dir/file';
+  let dirname = '/dir';
```
humphd: All these `let` keywords can be switched to `const`, since the variables aren't going to change value.

PR #697, opened 2019-01-29
Fix Issue 684: Replace var with const, let in filer/tests/spec/filer.buffer.spec.js

rscotchmer:
Replaced instances of var with const and let, depending on the variable being declared.  Const was used for the requires, whose values should not change.  Let was used for everything else, as the values were permutable and accessed only within the scope they were declared in.

### DIFF
--- a/tests/spec/filer.buffer.spec.js
+++ b/tests/spec/filer.buffer.spec.js
@@ -1,5 +1,7 @@
-var Filer = require('../../src');
-var expect = require('chai').expect;
+'use strict';
+
+const Filer = require('../../src');
+const expect = require('chai').expect;
 
 describe('Filer.Buffer', function() {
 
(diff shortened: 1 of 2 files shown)


ApolllonDev:
@humphd, @rscotchmer should `package-lock.json` be updated in PR?

tests/spec/filer.buffer.spec.js
```diff
@@ -12,38 +14,38 @@ describe('Filer.Buffer', function() {
   });
 
   it('should support .isBuffer()', function() {
-    var buf = Buffer.alloc(0);
+    let buf = Buffer.alloc(0);
```
humphd: This can actually be `const` too.  Since `buf` won't change at all in this scope, you can use `const` to signal that fact to JS.

Same for all the other cases of `let` below.  Go ahead and change them all to `const`.  Using `let` isn't wrong, but in this case it's actually more correct to say `const`.

tests/spec/filer.buffer.spec.js
```diff
@@ -12,38 +14,38 @@ describe('Filer.Buffer', function() {
...
-      var buf1 = new Buffer(s);
-      var buf2 = Buffer.from(s);
+      let s = 'Hello World';
+      let buf1 = new Buffer(s);
+      let buf2 = Buffer.from(s);
       expect(buf1).to.deep.equal(buf2);
     });
   });
-});
+});
```
humphd: In general we try to avoid this (i.e., not having a newline at the end of a file).  The reason we do this is so that if this file is concatenated with another file, the last line of this file, and the first line of some other file, don't get joined and cause some weird error.

Always include a newline at the end of all your files.

humphd:
@ApolllonDev yes, we can do that in a different PR.

PR #700, opened 2019-01-29
fixed Issue-689 indexedDB.js

vlogozzo:
I've made some changes to the indexedDB.js to use let and const on its local values


vlogozzo:
I updated my repo reverting the dependency changes, I saw that the pull request has my commit in it, are you notified when I do this, or should I leave a comment as well saying the changes have been made?



________________________________

Issue #699, opened 2019-01-29
Update code in filer/tests/spec/path-resolution.spec.js to replace var with const/let 


bonbon12:
var needs to be replaced with let and const with strict mode.

May i be assigned this task?

humphd:
@bonbon12 all yours, go for it.

PR #701, opened 2019-01-29
Issue-#699 FIX Updated File to use let/const and strict mode

bonbon12:
Replace var with let and const and added strict mode.

### DIFF
--- a/tests/spec/path-resolution.spec.js
+++ b/tests/spec/path-resolution.spec.js
@@ -1,16 +1,17 @@
-var Filer = require('../../src');
-var util = require('../lib/test-utils.js');
-var expect = require('chai').expect;
+'use strict';
+const Filer = require('../../src');
+const util = require('../lib/test-utils.js');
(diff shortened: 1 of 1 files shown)


ithompson4:
Alexei, I noticed 2 tests in the fs.open file failed for 1052.2 for OSX. Your code updates look fine. But go through your code again, make changes if necessary, and test it again. Use Git to add and commit all changes, push to origin to publish new local commits and create a new pull request.

Violet-XiaoWeiHuang:
Hello, 
I tested your unchanged code and the latest commit one. They both worked on my local terminal. I met the same issue as you. In your case, I may try to make a small change, like adding a new line. And then commit again. Sometimes it could be remote compiler issue. 
My another idea is [in your line 7](https://github.com/filerjs/filer/blob/f6dbb5844db564625e0f758c85ff95297fece9cf/tests/spec/path-resolution.spec.js#L7), it could be  'let', rather than 'const'. 
Nice work!

PR #705, opened 2019-01-29
Issue698

pbrahmbhatt3:
issue123: I have changed all the constants that were declared using var to const.


abdiG:
Awesome work. You seem to have left some variable as `var`. i recommend to use `let`

otkach:
Good job, but i'll suggest to add `'use strict';` to your files.

PR #706, opened 2019-01-29
Fixed #699 Replacing var with let/const and adding strict mode

### LINKED ISSUE #699: Update code in filer/tests/spec/path-resolution.spec.js to replace var with const/let

bonbon12:
Updating for better coding practices.

### DIFF
--- a/tests/spec/path-resolution.spec.js
+++ b/tests/spec/path-resolution.spec.js
@@ -1,16 +1,17 @@
-var Filer = require('../../src');
-var util = require('../lib/test-utils.js');
-var expect = require('chai').expect;
+'use strict';
+const Filer = require('../../src');
+const util = require('../lib/test-utils.js');
+const expect = require('chai').expect;
 
 // Support global URL and node's url module
-var URL = global.URL || require('url').URL;
+const URL = global.URL || require('url').URL;
 
 describe('path resolution', function() {
   beforeEach(util.setup);
   afterEach(util.cleanup);
 
   it('should follow a symbolic link to the root directory', function(done) {
-    var fs = util.fs();
+    let fs = util.fs();
 
     fs.symlink('/', '/mydirectorylink', function(error) {
       if(error) throw error;
@@ -19,7 +20,7 @@ describe('path resolution', function() {
         if(error) throw error;
 
         expect(result['node']).to.exist;
-        var _node = result['node'];
+        let _node = result['node'];
 
         fs.stat('/mydirectorylink', function(error, result) {
           expect(error).not.to.exist;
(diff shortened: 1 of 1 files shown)


tests/spec/path-resolution.spec.js
```diff
@@ -241,7 +242,7 @@ describe('path resolution', function() {
   });
 
   it('should properly remove trailing slashes with Path.removeTrailing()', function() {
-    var Path = Filer.Path;
+    let Path = Filer.Path;
```
hoaianhkhang:
```suggestion
    const Path = Filer.Path;
```
Great job with the code, I just think that since path wont be changing later on, you should put it const instead of let.

tests/spec/path-resolution.spec.js
```diff
@@ -232,7 +233,7 @@ describe('path resolution', function() {
   });
 
   it('should properly add trailing slashes with Path.addTrailing()', function() {
-    var Path = Filer.Path;
+    let Path = Filer.Path;
```
hoaianhkhang: I think this is the same with the previous one, the path should be a const because It will not be used later on

tests/spec/path-resolution.spec.js
```diff
@@ -1,16 +1,17 @@
...
-var util = require('../lib/test-utils.js');
-var expect = require('chai').expect;
+'use strict';
+const Filer = require('../../src');
+const util = require('../lib/test-utils.js');
+const expect = require('chai').expect;
 
 // Support global URL and node's url module
-var URL = global.URL || require('url').URL;
+let URL = global.URL || require('url').URL;
```
humphd: This can be `const`, it won't change.

tests/spec/path-resolution.spec.js
```diff
@@ -1,16 +1,17 @@
...
 
 describe('path resolution', function() {
   beforeEach(util.setup);
   afterEach(util.cleanup);
 
   it('should follow a symbolic link to the root directory', function(done) {
-    var fs = util.fs();
+    let fs = util.fs();
```
humphd: `const` here too, since `fs` will always be the same for this block.

tests/spec/path-resolution.spec.js
```diff
@@ -19,7 +20,7 @@ describe('path resolution', function() {
         if(error) throw error;
 
         expect(result['node']).to.exist;
-        var _node = result['node'];
+        let _node = result['node'];
```
humphd: Same thing here.  In fact, I don't think I see any `let` below that can't become a `const`.  Do you want to try changing them all, and see what eslint and the tests say?

PR #709, opened 2019-01-30
updated environment.js to use const insted of var for constant property

pbrahmbhatt3:
issue-698b fix: Updated the filer/src/shell/environment.js file to use const instead of var for a constant property

### DIFF
--- a/src/shell/environment.js
+++ b/src/shell/environment.js
@@ -1,3 +1,4 @@
+'use strict'; 
 const defaults = require('../constants.js').ENVIRONMENT;
 
 module.exports = function Environment(env) {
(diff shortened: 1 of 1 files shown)


src/shell/environment.js
```diff
@@ -1,4 +1,4 @@
-var defaults = require('../constants.js').ENVIRONMENT;
+const defaults = require('../constants.js').ENVIRONMENT;
```
vrozin: Hey, I think you should also include `'use strict';` to the head of the file.

pbrahmbhatt3: good catch! Thanks. :) @vrozin

Issue #703, opened 2019-01-29
Update code in filer/tests/spec/times.spec.js to use const/let syntax over var


dcoull:
The times.spec.js file uses the var syntax for variable declarations, as opposed to the more modern const/let syntax

PR #711, opened 2019-01-30
Addressed Issue #703

dcoull:
Updated variable declarations in times.spec.js from var to let and const. All tests passed


ndngo:
Some variables that were declared with let could be declared with const but this is more stylistic preference. Nice job!

tests/spec/times.spec.js
```diff
@@ -1,15 +1,15 @@
-var util = require('../lib/test-utils.js');
-var expect = require('chai').expect;
+const util = require('../lib/test-utils.js');
```
ithompson4: Add 'use strict'; at the top of your file to use strict mode.

humphd:
@dcoull how is this coming?

dcoull:
> @dcoull how is this coming?

Changes made as requested

humphd:
@dcoull this still didn't work, maybe we should try this together when we see each other tomorrow.  Can you bring your laptop and we do it together?

PR #712, opened 2019-01-30
Fixed #707 Replaced var with const and let and added strict mode

### LINKED ISSUE #707: Update "filer/src/fs-watcher.js" file to use const and let instead of var
@humphd I will be updating fs-watcher.js file to use const and let instead of var where applicable.

hjpatel16:


DavidLi119:
Hi hjpatel16, great edit of your file, however you forgot to change lines 14-16 from var to let/const. Also, though it may be redundant, please include a description of what you did to the code.

hjpatel16:
Thanks @DavidLi119, I was confused to use let for those remaining variables but I did make some changes

src/fs-watcher.js
```diff
@@ -46,12 +48,12 @@ function FSWatcher() {
       recursivePathPrefix = filename === '/' ? '/' : filename + '/';
     }
 
-    var intercom = Intercom.getInstance();
+    let intercom = Intercom.getInstance();
```
humphd: This could be `const`, since `intercom` won't change again.

src/fs-watcher.js
```diff
@@ -46,12 +48,12 @@ function FSWatcher() {
...
     intercom.on('change', onchange);
   };
 
   self.close = function() {
-    var intercom = Intercom.getInstance();
+    let intercom = Intercom.getInstance();
```
humphd: Same thing here, you can use `const`

Harsh2777:
Fixed #707 for fs-watcher.js file to use let and const instead of var where applicable. You can check under the files changed tab to review my changes

PR #713, opened 2019-01-30
Fixed #708: Added 'use strict' and replaced var with let/const in fs.appendFile.spec.js

### LINKED ISSUE #708: Updating filer/tests/spec/fs.appendFile.spec.js from var to let/const
fs.appendFile.spec.js has plenty of var declarations, may I update it to let/const declarations?

DavidLi119:
Added `'use strict'` at the top of the file, then replaced var with const:

```js
var ex = 'unchanging';
// change to
const ex = 'unchanging';
```
Same with `let`:

```js
var ex = run();
// change to
let ex = run();
```


tests/spec/fs.appendFile.spec.js
```diff
@@ -1,12 +1,14 @@
...
+const util = require('../lib/test-utils.js');
+const expect = require('chai').expect;
 
 describe('fs.appendFile', function() {
   const contents = 'This is a file.';
 
   beforeEach(function(done) {
     util.setup(function() {
-      var fs = util.fs();
+      let fs = util.fs();
```
humphd: This could probably be `const`

tests/spec/fs.appendFile.spec.js
```diff
@@ -16,13 +18,13 @@ describe('fs.appendFile', function() {
   afterEach(util.cleanup);
 
   it('should be a function', function() {
-    var fs = util.fs();
+    let fs = util.fs();
```
humphd:
```suggestion
    const fs = util.fs();
```

DavidLi119:
Made all changes that were requested professor. Also, I didn't realize that committing via VSC would commit to github as well, hence the 1 letter commit name at first. I reverted and recommitted properly.

humphd:
@DavidLi119 I see you've pushed a new commit, and wanted to fix your commit message.  Here's a tip:

```
git commit --amend -m "New commit message"
```

The `--amend` will add/to, update, etc. the old commit with whatever you add, which can be as simple as the commit message changing.

DavidLi119:
Ah, thanks for the recommendation, I will keep it in mind for next time.

PR #717, opened 2019-01-30
Issue-686 - rfs.mkdtemp.spec.js

cmchumak:


tests/spec/fs.mkdtemp.spec.js
```diff
@@ -1,6 +1,8 @@
 const util = require('../lib/test-utils.js');
 const expect = require('chai').expect;
 
+'use strict';
```
bonbon12: 'use strict' should be at the top of the code to ensure all the code below is ran in strict mode

cmchumak: Done. Thanks.

Issue #664, opened 2019-01-23
Update and modernize /tests/spec/errors.spec.js to use let, const, and strict mode for best practices.


PrimeN:
Moderin java script prefers the use of let, const, and strict mode for best practices.

humphd:
Go for it @PrimeN.

PR #718, opened 2019-01-30
Update errors.spec.js

PrimeN:
Updating errors.spec.js to use const/let appropriately instead of var.

### DIFF
--- a/tests/spec/errors.spec.js
+++ b/tests/spec/errors.spec.js
@@ -1,4 +1,4 @@
-'using strict';
+'use strict';
 
 const Filer = require('../../src');
 const expect = require('chai').expect;
@@ -137,7 +137,7 @@ describe('Filer.Errors', function() {
   });
 
   it('should include all expected properties by default', function() {
-    let err = new Filer.Errors.ENOENT();
+    const err = new Filer.Errors.ENOENT();
     expect(err.name).to.equal('ENOENT');
     expect(err.code).to.equal('ENOENT');
     expect(err.errno).to.equal(34);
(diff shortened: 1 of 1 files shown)


tests/spec/errors.spec.js
```diff
@@ -1,5 +1,7 @@
-var Filer = require('../../src');
-var expect = require('chai').expect;
+'using strict';
```
humphd: You should change `using` to `use`

tests/spec/errors.spec.js
```diff
@@ -135,15 +137,15 @@ describe('Filer.Errors', function() {
   });
 
   it('should include all expected properties by default', function() {
-    var err = new Filer.Errors.ENOENT();
+    let err = new Filer.Errors.ENOENT();
```
humphd: This could actually be `const` vs. `let`, since `err` will never change during the function scope.

tests/spec/errors.spec.js
```diff
@@ -135,15 +137,15 @@ describe('Filer.Errors', function() {
...
     expect(err.name).to.equal('ENOENT');
     expect(err.code).to.equal('ENOENT');
     expect(err.errno).to.equal(34);
     expect(err.message).to.equal('no such file or directory');
   });
 
   it('should include extra properties when provided', function() {
-    var err = new Filer.Errors.ENOENT('This is the message', '/this/is/the/path');
+    let err = new Filer.Errors.ENOENT('This is the message', '/this/is/the/path');
```
humphd: Same thing here.

humphd:
@PrimeN one tip: you're fixing issue #664, so in your PR you can mention it by saying `Fixes #664` and then GitHub will auto-close your issue when the PR gets merged.  Next time!

Issue #672, opened 2019-01-25
Refactor "filer/tests/spec/fs.mkdir.spec.js" to follow ES6 syntax with "const/let" instead of "var"


ApolllonDev:
### `fs.mkdir.spec.js` file contains `var` declarations 

In order to follow `ES6` standards, the following changes are required:

- [x] Refactor the code to use `let` and `const` instead of `var`
- [x] Invoke a `strict mode`

ApolllonDev:
#722

PR #722, opened 2019-01-30
Fix issue #672 : Refactor "filer/tests/spec/fs.mkdir.spec.js" to follow ES6 syntax with "const/let" instead of "var"

ApolllonDev:
- [x] Invoked a `strict mode`
- [x] Refactored the code to use ~~let~~ and `const` instead of `var`

### DIFF
--- a/tests/spec/fs.mkdir.spec.js
+++ b/tests/spec/fs.mkdir.spec.js
@@ -12,7 +12,6 @@ describe('fs.mkdir', function () {
   });
 
   it('should return an error if part of the parent path does not exist', function (done) {
-    'use strict';
     const fs = util.fs();
 
     fs.mkdir('/tmp/mydir', function (error) {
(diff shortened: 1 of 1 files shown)


tests/spec/fs.mkdir.spec.js
```diff
@@ -88,7 +89,7 @@ describe('fs.promises.mkdir', function () {
   });
 
   it('should return a promise', function () {
-    var fs = util.fs();
+    const fs = util.fs();
     expect(fs.promises.mkdir('/tmp')).to.be.a('promise');
   });
 });
```
humphd: One tip: it's a good idea to always include a newline at the end of all your files (I know this isn't your mistake, it was already like this).  We do this, and GitHub shows its absence, because we want to avoid having concatenated files break when the end of one file and the start of another get joined.  Just a tip for next time.

ApolllonDev: Sounds good, will do.

PR #723, opened 2019-01-30
fix var to const

QingLianLiu:
Fix var into const


PriyankaCodes99:
Looks good. As stated, 'use strict' can be added for strict mode.

humphd:
@QingLianLiu how is this coming?

Issue #702, opened 2019-01-29
Code Update in filer/tests/spec/filer.spec.js (var to const/let)


SahibArora:
I would like to update the old declarations in filer.spec.js
I will change the var to const for all the variables being used as objects of the libraries and for any other local variables used in the file will be changed to let.
If nobody is doing so already, I would like to take this opportunity and update the file.

PR #726, opened 2019-01-30
'Use strict' @top and updated the file with const/let  [issue-702]

SahibArora:
Added strict mode in filer.spec.js
 Learn more about strict > https://www.w3schools.com/js/js_strict.asp

```JS
var abs = "..";

// changed to 

const abs = "..";
// Variables which are objects of library are changed to const and rest are changed to let.
```
Please advice if I missed something.
Thank you

### DIFF
--- a/tests/spec/filer.spec.js
+++ b/tests/spec/filer.spec.js
@@ -1,6 +1,7 @@
-var Filer = require('../../src');
-var util = require('../lib/test-utils');
-var expect = require('chai').expect;
+'use strict';
+const Filer = require('../../src');
+const util = require('../lib/test-utils');
+const expect = require('chai').expect;
 
 describe('Filer', function() {
   it('is defined', function() {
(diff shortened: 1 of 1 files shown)


tests/spec/filer.spec.js
```diff
@@ -45,21 +46,21 @@ describe('Filer', function() {
   });
 
   it('must honor the \'FORMAT\' flag', function(done) {
-    var name = 'local-test';
+    let name = 'local-test';
```
pbrahmbhatt3: Hi @SahibArora  
```suggestion
    const name = 'local-test';
```
I think you can update the name variable to be a const as it is getting reassigned in the code.

SahibArora: Hi @pbrahmbhatt3,
Thanks for coming and looking over.
I will do the necessary changes.

humphd:
One tip @SahibArora: you use `[issue-702]` in your issue above, but it would be better to say `Fixes #702`, which GitHub will use to auto-close your bug when this PR lands.  Next time!

Issue #715, opened 2019-01-30
Update filer 'filer/tests/spec/fs.lstat.spec.js' to use const and let instead of var


hoaianhkhang:
Hi, I may I have the authority to fix some of the var mistakes in this file into using const or let in appropriate situation?

hoaianhkhang:
Fixes #715

PR #727, opened 2019-01-30
Issue #715 Update fs.lstat.spec.js to have proper const and let instead of var and added 'use strict'

hoaianhkhang:
I edit the code to change some let into const and tested it again

### DIFF
--- a/tests/spec/fs.lstat.spec.js
+++ b/tests/spec/fs.lstat.spec.js
@@ -1,74 +1,50 @@
-var util = require('../lib/test-utils.js');
-var expect = require('chai').expect;
+"use strict";
+const util = require('../lib/test-utils.js');
+const expect = require('chai').expect;
 
-describe('fs.lstat', function() {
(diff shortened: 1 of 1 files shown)


hoaianhkhang:
I have changed the file, can you check it again please?

tests/spec/fs.lstat.spec.js
```diff
@@ -13,7 +13,6 @@ describe('fs.lstat', function() {
 
   it('should return an error if path does not exist', function(done) {
     const fs = util.fs();
-
```
humphd: Why remove all these blank lines?  It's normally a bad idea to touch lines that aren't really part of this fix.  The reason it's generally avoided is that it will alter the history of these lines in git, making them look like they were changed in this PR, but really they weren't.

hoaianhkhang: I have changed the file again with the space

PR #729, opened 2019-01-30
Issue 710 fixed

abdiG:
In file `filer/tests/spec/fs.exists.spec.js` i changed `var` to `let` and added `'use strick'`


kwchan19:
I'm seeing that you used 'let' for all of the variables. It would be better to use 'const', as the variables do not change values.

tests/spec/fs.exists.spec.js
```diff
@@ -1,17 +1,18 @@
...
+let util = require('../lib/test-utils.js');
+let expect = require('chai').expect;
 
 describe('fs.exists', function() {
   beforeEach(util.setup);
   afterEach(util.cleanup);
 
   it('should be a function', function() {
-    var fs = util.fs();
+    let fs = util.fs();
```
humphd: This can get changed to `const` since `fs` won't change in the current scope.

tests/spec/fs.exists.spec.js
```diff
@@ -1,17 +1,18 @@
...
     expect(typeof fs.exists).to.equal('function');
   });
 
   it('should return false if path does not exist', function(done) {
-    var fs = util.fs();
+    let fs = util.fs();
```
humphd: This can get changed to `const` since `fs` won't change in the current scope.

tests/spec/fs.exists.spec.js
```diff
@@ -20,7 +21,7 @@ describe('fs.exists', function() {
   });
 
   it('should return true if path exists', function(done) {
-    var fs = util.fs();
+    let fs = util.fs();
```
humphd: This can get changed to `const` since `fs` won't change in the current scope.

tests/spec/fs.exists.spec.js
```diff
@@ -37,7 +38,7 @@ describe('fs.exists', function() {
   });
 
   it('should follow symbolic links and return true for the resulting path', function(done) {
-    var fs = util.fs();
+    let fs = util.fs();
```
humphd: This can get changed to `const` since `fs` won't change in the current scope.

tests/spec/fs.exists.spec.js
```diff
@@ -58,7 +59,7 @@ describe('fs.exists', function() {
   });
   
   it('should follow symbolic links and return false if for the resulting path does not exist', function(done) {
-    var fs = util.fs();
+    let fs = util.fs();
```
humphd: This can get changed to `const` since `fs` won't change in the current scope.

humphd:
@abdiG how is this coming?

humphd:
@abdiG how is this coming?

abdiG:
@humphd I have made all the changes you asked. And sorry it took a while.

humphd:
Fix and merged in https://github.com/filerjs/filer/commit/3d10d6431c853d69006a42257620b818e875b8d1

PR #730, opened 2019-01-31
Issue 728: Updated 'filer/tests/spec/fs.shell.spec.js' to use 'const/let' instead of 'var'

kwchan19:
- Added strict mode in 'filer/tests/spec/fs.shell.spec.js'
- Changed 'var' variables to use 'const'


humphd:
@kwchan19 @FineHHDocumentation, I'm confused about this PR.  @kwchan19 opens it, and then there are 6 commits from @FineHHDocumentation.  What's going on?

PR #731, opened 2019-01-31
Issue#719: Updated var to const/let

ael-mas:
Updated var declarations to use const or let instead.
Added strict mode to file.

### DIFF
--- a/tests/spec/fs.read.spec.js
+++ b/tests/spec/fs.read.spec.js
@@ -1,19 +1,20 @@
-var util = require('../lib/test-utils.js');
-var expect = require('chai').expect;
+'use strict';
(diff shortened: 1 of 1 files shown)


humphd:
@ael-mas how is this coming?  Any update?

humphd:
@ael-mas did you want this PR closed in favour of the new one you opened?

ael-mas:
The new one should have the correct variable declarations using const instead of let. Apologies if I didn't modify my original changes properly on github.

humphd:
OK, no problem.  Closing in favour of #741.  In future, you can just make more commits and push to your same branch, and it will add to/update an existing PR vs. making a new one.

Issue #714, opened 2019-01-30
Update file 'filer/tests/spec/fs.writeFile-readFile.spec' to use let/const instead of var


otkach:
There are a lot of 'var' in file 'filer/tests/spec/fs.writeFile-readFile.spec' which need to be changed to const or let.
@humphd can i work on this file?

PR #732, opened 2019-01-31
Updated file for issue #714

otkach:
All `vars` were changed to `let` or `const`. Also `'use strict';` was added to file.


otkach:
@humphd all changes was made due to your request. Please check if everything is fine.

PR #735, opened 2019-02-01
Changed the declaration types [issue-734]

aqeelparpia:
Fixed!

### DIFF
--- a/tests/spec/fs.xattr.spec.js
+++ b/tests/spec/fs.xattr.spec.js
@@ -201,7 +201,7 @@ describe('fs.xattr', function() {
 
         fs.removexattr('/testfile', 'testenoattr', function (error) {
           expect(error).to.exist;
-          expect(error.code).to.equal('ENOATTR');gi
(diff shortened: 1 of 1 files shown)


tests/spec/fs.xattr.spec.js
```diff
@@ -1,12 +1,12 @@
-var util = require('../lib/test-utils.js');
-var expect = require('chai').expect;
+const util = require('../lib/test-utils.js');
```
humphd: Can you add `'use strict';` to the top of this file please?

tests/spec/fs.xattr.spec.js
```diff
@@ -170,8 +170,8 @@ describe('fs.xattr', function() {
   });
 
   it('should set and get an extended attribute of a path', function(done) {
-    var fs = util.fs();
-    var name = 'test';
+    const fs = util.fs();
+    let name = 'test';
```
humphd: This can be `const`

PR #740, opened 2019-02-08
fs.lstat should return a Stats object including mode

humphd:
Working on my Linux-in-Browser shell again, and hit a bug where `fs.stat` is returning a proper `Stats` object with a `mode`, but `fs.lstat` is returning a raw database object.  This creates a proper `Node` object in `lstat`.  I've also added tests.


humphd:
Reduced test coverage is due to error handling paths that we can't really test, ignorable.

PR #745, opened 2019-02-26
add test for fs.open with wx flag for existing file

### LINKED ISSUE #623: Add test for fs.open() with O_WRITE + O_EXCLUSIVE and existing file
https://codecov.io/gh/filerjs/filer/src/master/src/filesystem/implementation.js#L594 is a branch we don't test.  To hit it, I think we need:

* add a test to `fs.open.spec.js`
* `mkdir` a directory `/tmp` then create a file in the directory, `/tmp/file`
* now try to `fs.open()` that path and use the `'wx'` flag
* expect an `ENOENT` error to be returned

Somewhat similar to the structure of [this test](https://github.com/filerjs/filer/blob/master/tests/spec/fs.open.spec.js#L50).

cdrani:
Fixes #623

### DIFF
--- a/src/filesystem/implementation.js
+++ b/src/filesystem/implementation.js
@@ -620,7 +620,7 @@ function open_file(context, path, flags, mode, callback) {
       directoryData = result;
       if(directoryData.hasOwnProperty(name)) {
         if(flags.includes(O_EXCLUSIVE)) {
-          callback(new Errors.ENOENT('O_CREATE and O_EXCLUSIVE are set, and the named file exists', path));
+          callback(new Errors.EEXIST('O_CREATE and O_EXCLUSIVE are set, and the named file exists', path));
         } else {
           directoryEntry = directoryData[name];
           if(directoryEntry.type === NODE_TYPE_DIRECTORY && flags.includes(O_WRITE)) {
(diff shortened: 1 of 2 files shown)


cdrani:
Sure, no problem.

cdrani:
The test is failing, expecting 'EEXIST' to equal 'ENOENT'.

humphd:
You can just run `npm test` and it will rebuild automatically.  Did you change your test file to expect `EEXIST` after making that change to the implementation?

Also, can you push your changes up again and I'll review again, and help fix anything.

cdrani:
Yes, I made all the changes. Will push now.

cdrani:
Concerning this
>  No binary for FirefoxHeadless browser on your platform.
  Please, set "FIREFOX_BIN" env variable.

I noticed that I have ChromeHeadless, but not FirefoxHeadless. How are these setup? Why do I have one binary and not the other?

humphd:
@cdrani do you have Firefox installed on your machine?  That should be enough to make this work.

At any rate, this looks great, nice work.  I'm going to merge.

cdrani:
Oh, I use the developer version. That might be the issue. Thanks for the help!

PR #746, opened 2019-03-06
Issue 627

### LINKED ISSUE #627: Add tests for fs.write to increase coverage
We have a bunch of missed branches in our `fs.write` code: https://codecov.io/gh/filerjs/filer/src/master/src/filesystem/implementation.js#L1833.  To improve it, we need:

* test that tries to use a non-existent `fd`
* test that tries to open a file without the `O_WRITE` flag and then call `fs.write`
* test that tries to write a `buffer` whose length is too small (`buffer.length - offset < length`)

rscotchmer:
Fixes #627 by adding 3 tests to fs.write():

- Test that tries to use a non-existent fd
- Test that tries to open a file without the O_WRITE flag and then call fs.write
- Test that tries to write a buffer whose length is too small (buffer.length - offset < length)

### DIFF
--- a/tests/spec/fs.write.spec.js
+++ b/tests/spec/fs.write.spec.js
@@ -72,12 +72,12 @@ describe('fs.write', function() {
 
     fs.write('/myfile', buffer, 0, buffer.length, 0, function(error) {
       expect(error).to.exist;
-      expect(undefined).to.equal(undefined);
+      expect(error.code).to.equal('EBADF');
       done();
     });
   });
 
-  it('should fail to write a buffer whose length is too small', function(done) {
+  it('should fail when trying to write more bytes than are available in the buffer (length too long)', function(done) {
     const fs = util.fs();
     const buffer = Buffer.from([1, 2, 3, 4, 5, 6, 7, 8]);
 
@@ -86,7 +86,7 @@ describe('fs.write', function() {
 
       fs.write(fd, buffer, 0, (buffer.length + 10), 0, function(error) {
         expect(error).to.exist;
-        expect(undefined).to.equal(undefined);
+        expect(error.code).to.equal('EIO');
         fs.close(fd, done);
       });
     });
--- a/tests/spec/fs.write.spec.js
+++ b/tests/spec/fs.write.spec.js
@@ -65,4 +65,23 @@ describe('fs.write', function() {
       });
     });
   });
+
+  it('should fail to write data to a file opened without the O_WRITE flag', function(done) {
+    const fs = util.fs();
+    const buffer = Buffer.from([1, 2, 3, 4, 5, 6, 7, 8]);
+
+    fs.mknod('/myfile', 'FILE', function(err) {
+      if (err) throw err;
+
+      fs.open('/myfile', 'r', function(error, fd) {
+        if(error) throw error;
+  
+        fs.write(fd, buffer, 0, buffer.length, 0, function(error) {
+          expect(error).to.exist;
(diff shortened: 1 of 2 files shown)


tests/spec/fs.write.spec.js
```diff
@@ -65,4 +65,49 @@ describe('fs.write', function() {
...
     });
   });
+
+  it('should fail to write to a file that does not exist', function(done) {
+    const fs = util.fs();
+    const buffer = Buffer.from([1, 2, 3, 4, 5, 6, 7, 8]);
+
+    fs.write('/myfile', buffer, 0, buffer.length, 0, function(error) {
+      expect(error).to.exist;
+      expect(undefined).to.equal(undefined);
```
humphd: This assertion doesn't really add much.  What if we modify it to check the type of error we get back?  For example:

```
```

tests/spec/fs.write.spec.js
```diff
@@ -65,4 +65,49 @@ describe('fs.write', function() {
...
+      done();
+    });
+  });
+
+  it('should fail to write a buffer whose length is too small', function(done) {
```
humphd: Let's clarify this slightly:

```js
it('should fail when trying to write more bytes than are available in the buffer (length too long)', ...`)
```

tests/spec/fs.write.spec.js
```diff
@@ -65,4 +65,49 @@ describe('fs.write', function() {
...
+    const fs = util.fs();
+    const buffer = Buffer.from([1, 2, 3, 4, 5, 6, 7, 8]);
+
+    fs.open('/myfile', 'w', function(error, fd) {
+      if(error) throw error;
+
+      fs.write(fd, buffer, 0, (buffer.length + 10), 0, function(error) {
+        expect(error).to.exist;
+        expect(undefined).to.equal(undefined);
```
humphd: Again, let's check `error.code` for the expected type here.

tests/spec/fs.write.spec.js
```diff
@@ -65,4 +65,49 @@ describe('fs.write', function() {
...
+
+    fs.mknod('/myfile', 'FILE', function(err) {
+      if (err) throw err;
+
+      fs.open('/myfile', 'r', function(error, fd) {
+        if(error) throw error;
+  
+        fs.write(fd, buffer, 0, buffer.length, 0, function(error) {
+          expect(error).to.exist;
+          expect(error.code).to.equal('EBADF');
```
humphd: What you've done here is better.

rscotchmer:
Made the changes -- hope these are good to go!  Thanks for reviewing my work!

tests/spec/fs.write.spec.js
```diff
@@ -65,4 +65,49 @@ describe('fs.write', function() {
       });
     });
   });
+
+  it('should fail to write to a file that does not exist', function(done) {
+    const fs = util.fs();
+    const buffer = Buffer.from([1, 2, 3, 4, 5, 6, 7, 8]);
+
+    fs.write('/myfile', buffer, 0, buffer.length, 0, function(error) {
```
humphd: `fs.write` takes an `fd`, so passing it a filepath as a `string` is really testing what we do when we don't pass an `fd`.  If I do this in node, it throws.  Probably returning `EBADF` is OK here, since it literally is a bad file descriptor.

That said, let's adjust the message on line 69 to be more clear: `it('should fail if given a file path vs. an fd'`

Issue #620, opened 2018-12-11
Implement lchown


humphd:
https://nodejs.org/api/fs.html#fs_fs_lchown_path_uid_gid_callback

andrewkoung:
Could I try learning this and implementing it?

humphd:
@andrewkoung sure, let's discuss in class tomorrow.

andrewkoung:
@humphd I had forgotten to ask this in class about context.getObject(). Let's say this line of code `context.getObject(superNode.rnode, check_root_directory_node);` fires, does it look through the key-value table using `superNode.rnode` as a key and passes the value into the method `check_root_directory_node`?

humphd:
@andrewkoung yes, that is exactly what it does.  It basically means: "get the file node for the super-node, and once we have that file node object, pass it to the check_root_directory_node function callback".

PR #752, opened 2019-04-07
Fix #749, #620: implement lchown()

### LINKED ISSUE #749: Refactor find_node() to have a parameter to make deferencing optional
Currently the find_node() method in implementation.js will attempt to dereference to find the root node. It would be a good idea to have an option to not dereference for situations where you need to work on symlinks.

andrewkoung:
Let me know if my logic was wrong, but I believe this was the correct flow.

### DIFF
--- a/src/filesystem/implementation.js
+++ b/src/filesystem/implementation.js
@@ -278,6 +278,14 @@ function find_node(context, path, callback) {
   }
 }
 
+/**
+ * find_lnode
+ */
+// in: file or directory path
+// out: node structure, or error
+function find_symbolic_node(context, path, callback) {
+  lstat_file(context, path, callback);
+} 
 
 /**
  * set extended attribute (refactor)
@@ -2060,6 +2068,19 @@ function fchown_file(context, ofd, uid, gid, callback) {
   ofd.getNode(context, update_owner);
 }
 
+function lchown_file(context, path, uid, gid, callback) { 
+  function update_owner(error, node) {
+    if (error) {
+      callback(error);
+    } else {
+      node.uid = uid;
+      node.gid = gid;
+      update_node_times(context, path, node, { mtime: Date.now() }, callback);
+    }
+  }
+  find_symbolic_node(context, path, update_owner);
+} 
+
 function getxattr(context, path, name, callback) {
   getxattr_file(context, path, name, callback);
 }
@@ -2244,6 +2265,17 @@ function fchown(context, fd, uid, gid, callback) {
   }
 }
 
+function lchown(context, path, uid, gid, callback) {
+  if(!isUint32(uid)) {
+    return callback(new Errors.EINVAL('uid must be a valid integer', uid));
+  }
+  if(!isUint32(gid)) {
+    return callback(new Errors.EINVAL('gid must be a valid integer', gid));
+  }
+
+  lchown_file(context, path, uid, gid, callback);
+}
+
 function rename(context, oldpath, newpath, callback) {
   oldpath = normalize(oldpath);
   newpath = normalize(newpath);
@@ -2425,7 +2457,7 @@ module.exports = {
   ftruncate,
   futimes,
   getxattr,
-  // lchown - https://github.com/filerjs/filer/issues/620
+  lchown,
   // lchmod - https://github.com/filerjs/filer/issues/619
   link,
   lseek,
--- a/src/filesystem/implementation.js
+++ b/src/filesystem/implementation.js
@@ -943,47 +943,7 @@ function fstat_file(context, ofd, callback) {
 }
 
 function lstat_file(context, path, callback) {
-  path = normalize(path);
-  var name = basename(path);
-  var parentPath = dirname(path);
-
-  var directoryNode;
-  var directoryData;
-
-  if(ROOT_DIRECTORY_NAME === name) {
-    find_node(context, path, callback);
-  } else {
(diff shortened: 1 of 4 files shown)


src/filesystem/implementation.js
```diff
@@ -278,6 +278,14 @@ function find_node(context, path, callback) {
   }
 }
 
+/**
+ * find_lnode
+ */
+// in: file or directory path
+// out: node structure, or error
+function find_symbolic_node(context, path, callback) {
```
modeswitch: This should be `find_symlink_node` instead of `find_symbolic_node`.

src/filesystem/implementation.js
```diff
@@ -278,6 +278,14 @@ function find_node(context, path, callback) {
...
+  lstat_file(context, path, callback);
```
modeswitch: I think you want to move the implementation from `lstat_file` into this function and re-implement `lstat_file` using this function instead.

src/filesystem/implementation.js
```diff
@@ -2060,6 +2068,19 @@ function fchown_file(context, ofd, uid, gid, callback) {
...
+  function update_owner(error, node) {
+    if (error) {
+      callback(error);
+    } else {
+      node.uid = uid;
+      node.gid = gid;
+      update_node_times(context, path, node, { mtime: Date.now() }, callback);
+    }
+  }
+  find_symbolic_node(context, path, update_owner);
```
modeswitch: Fix this when you rename the function to `find_symlink_node`.

src/filesystem/interface.js
```diff
@@ -340,7 +340,7 @@ function FileSystem(options, callback) {
     { name: 'ftruncate' },
     { name: 'futimes' },
     { name: 'getxattr', promises: true, absPathArgs: [0] },
-    // lchown - https://github.com/filerjs/filer/issues/620
+    { name: 'lchown' },
     // lchmod - https://github.com/filerjs/filer/issues/619
```
modeswitch: You can remove this commented line, it's no longer relevant.

src/filesystem/implementation.js
```diff
@@ -278,6 +278,14 @@ function find_node(context, path, callback) {
   }
 }
 
+/**
+ * find_lnode
```
modeswitch: Update this comment to reflect the correct function name.

modeswitch (changes requested):
Quick first pass. @humphd I haven't reviewed the implementation itself.

src/filesystem/implementation.js
```diff
@@ -279,12 +279,52 @@ function find_node(context, path, callback) {
...
+  var name = basename(path);
+  var parentPath = dirname(path);
+
+  var directoryNode;
+  var directoryData;
+
+  if(ROOT_DIRECTORY_NAME === name) {
+    find_node(context, path, callback);
+  } else {
+    find_node(context, parentPath, read_directory_data);
```
humphd: Question: won't this just undo your work here, and give back a de-referenced symlink via `find_node`?  I thought the goal here was to get back the symlink vs. linked file node.

andrewkoung: That seems to make sense and I missed that, but lstat_file seems to be using this approach?

andrewkoung: What is the difference between stat and node really?

humphd: `find_node` and `stat` are basically identical: they both get a File Node object from the database for a given path.

I think I'm wrong, looking at `lstat_file`:  https://github.com/filerjs/filer/blob/master/src/filesystem/implementation.js#L897-L939.  It's only getting the base dir, and then using that to read out the directory entries.  A directory node's data is just an array of filenames stored in the dir.

It's using `find_node` just like this.  I assume you've pulled your code from this?  If so, can you also change `lstat_file` to just call your new function?

humphd:
Updated the title of this PR, since it will fix #749, #620.

src/filesystem/interface.js
```diff
@@ -340,8 +340,7 @@ function FileSystem(options, callback) {
     { name: 'ftruncate' },
     { name: 'futimes' },
     { name: 'getxattr', promises: true, absPathArgs: [0] },
-    // lchown - https://github.com/filerjs/filer/issues/620
-    // lchmod - https://github.com/filerjs/filer/issues/619
+    { name: 'lchown' },
```
humphd: This should be `{ name: 'lchown', promises: true },` so that it automatically adds a `.promises.lchown` version for us.

humphd (changes requested):
This is looking really good.  One thing it lacks is any kind of testing for the central difference between this and `chown`, namely, tests that work with symlinks and make sure that when you all `chown` it changes the referenced file, and when you call `lchown`, it changes the symlink node instead.  I'd add some tests that basically do this:

- create a file named `/file`
- create a symlink to `/file` called `/link`
- `chown` /link` to 500 500 or something
- `lchown` /link to 600 600 or something (different from the first)
- call `stat` on `/link` and expect the uid/gid to match 500
- call `lstat` on `/link` and expect the uid/gid to match 600

I'd break all that up into two separate tests, one for `stat`/`chown` and the other for `lstat`/`lchown`.  This way we'll know that they work as expected on the proper node.

Great work on this.

andrewkoung:
There happens to be a test for chown() already, but I've created the test for lchown()!

Issue #747, opened 2019-03-07
flags is not valid


tinchoz49:
Hi everyone!

I'm trying to do this and I'm getting flags is not valid:

```js
fs.open("/file.txt", fs.constants.O_RDWR | fs.constants.O_CREAT, function(err, fd) {
  if(err) throw err;
  fs.fstat(fd, function(err, stats) {
    if(err) throw err;
    fs.close(fd);
  });
});
```

humphd:
Hey @tinchoz49, this is a bug, thanks for filing.  We currently assume flags are in string form (`'w+'` is closest to what you'd want), and don't yet support passing flags as a bitwise number.  We should.

Fixing this would involve making our code do what node does:

* [docs](https://nodejs.org/api/fs.html#fs_fs_open_path_flags_mode_callback) where it discusses `flags` being a `String` or `Number`
* [code for `stringToFlags()` in node](https://github.com/nodejs/node/blob/3c277d7e92adf1464e41d27677fcdc05fec41a2b/lib/internal/fs/utils.js#L325-L360)

A simple fix would be to alter our [`O_FLAGS` object](https://github.com/filerjs/filer/blob/master/src/constants.js#L54-L65) to use numeric keys vs. strings, and convert strings to numbers like node does.

humphd:
@rscotchmer not sure what you're looking to do for your 0.3 bugs, but this one might be interesting to you.  I could mentor you through it, if you have interest.  No pressure if you have other things you want to try next.

rscotchmer:
@humphd I'd love to work on this!  Thanks for notifying me!

humphd:
@rscotchmer great, here are some more thoughts.

We currently use strings for flags, and we should switch to using numbers.  If a string is passed to us, we should convert it into a number.  See the code I linked to above from node, which does what we want.

Next, we do a bunch of checks all over like this: https://github.com/filerjs/filer/blob/master/src/filesystem/implementation.js#L622.  Instead of doing this, we'd want to convert that to use the `flags` number, and bitwise `&` (and) it with the constant we want to check against.  The constants we want are defined in https://github.com/filerjs/filer/blob/master/src/constants.js#L88.

We could start by converting the code in the first comment above into a test we add to the `fs.open.spec.js` test file, and add more as we go.  I have no doubt you'll find other bugs while fixing this, since we probably don't do the right thing in all cases with flags vs. what node does.

[...]

PR #753, opened 2019-04-10
Fix Issue 747

CDNRae:
Fixes issue #747 by allowing fs.open flags to be in the form of a string or bitwise number.  Adds a single test for creating a file using `fsConstants.O_CREAT`.


src/filesystem/implementation.js
```diff
@@ -1618,7 +1612,43 @@ function fremovexattr_file (context, ofd, name, callback) {
...
+  }
+  else if(flags & stringToFlags('ax')) {
+    return flags;
+  }
+  else if(flags & stringToFlags('ax+')) {
+    return flags;
+  }
+
+  return null;
+
```
humphd: Let's remove the blank lines and comment below.

src/filesystem/implementation.js
```diff
@@ -1633,6 +1663,7 @@ function validate_file_options(options, enc, fileMode){
 }
 
 function open(context, path, flags, mode, callback) {
+  flags = stringToFlags(flags);
```
humphd: Let's move this down to be with the other flags validation happening on 1691

humphd (changes requested):
Nice work on this.  After I get my grades submitted, I'll spend some more time on this, but in the meantime, I've left some feedback.  I think this is pretty close to be ready.

One thing we could do to shorten the code in `implementation.js` would be to do the following at the top after `Constants` is required (~line 8):

```js
const {
  O_RDONLY,
  O_WRONLY,
  O_RDWR,
  O_CREAT,
  O_TRUNC,
  O_APPEND
} = Constants.fsContansts;
```

Then we can get rid of all the `Constants.fsConstants.*` when you just want to use something like `O_RDONLY`.  Make sense?

src/filesystem/implementation.js
rscotchmer: I ended up moving that line, along with the validation, up a little further in the function -- `flags` gets used prior to line 1691, but until this most recent change, wasn't actually validated until that line.  I figured it would make more sense to validate them almost immediately, or at least do that before they're used.  Thank you for pointing this out, though -- I don't think I would have caught that mistake if it weren't for this suggestion!

humphd (approved):
I think this looks good, excellent work.  I don't have time to test it today, but will try to do so in the next few days, and can likely land it then.  The code seems right to me.

Thanks for finishing this!

PR #763, opened 2021-02-20
Update deps, add GitHub Actions CI

humphd:
Updating all our deps and moving to GitHub Actions for CI.  I'll probably need to do a few commits to get this action to work correctly with our headless browsers.


humphd:
@modeswitch I think you might have to enable the GitHub actions setting globally for the FilerJS org and/or make me an admin so I can.

humphd:
Gonna merge this and continuing iterating to get this working.

PR #768, opened 2021-03-20
docs: correct error in buffer shim documentation

bcheidemann:
This PR corrects an error in the Buffer shim documentation. Using webpack.ProvidePlugin as previously documented would result in Buffer being globally defined as:
`{ __esModule: true, default: ..., Buffer: ... }`
The update to the documentation results in Buffer being resolved correctly by webpack.

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,7 +187,7 @@ module.exports = {
   },
   plugins: [
     new webpack.ProvidePlugin({
-      Buffer: 'buffer',
+      Buffer: ['buffer', 'Buffer'],
     }),
   ]
 }


humphd:
I still think a test that uses webpack to do the build would be smart, since all of these shims are aimed at using it in that context, otherwise regressions are going to happen where we break those builds and expectations.

humphd (approved):

bcheidemann:
> I still think a test that uses webpack to do the build would be smart, since all of these shims are aimed at using it in that context, otherwise regressions are going to happen where we break those builds and expectations.

Something like this? https://github.com/filerjs/filer/pull/770

PR #770, opened 2021-03-21
Build and run tests with webpack

bcheidemann:


bcheidemann:
@humphd Is this the kind of thing you were talking about?

tests/webpack.config.js
```diff
@@ -0,0 +1,61 @@
...
+const NodePolyfillPlugin = require('node-polyfill-webpack-plugin');
+const MiniCssExtractPlugin = require('mini-css-extract-plugin');
+const path = require('path');
+
+module.exports = env => ({
+  mode: 'development',
+  entry: path.resolve(__dirname, './webpack-tests.js'),
+  resolve: {
+    alias: {
+      'fsProvider': path.resolve(__dirname, '../shims/providers/default'),
```
humphd, line 12: Forgive my webpack ignorance here: can we not use Filer in this setup the way you're outlining in the README?  It would be great if we had test code that did what the README suggests, so we know if it breaks.

humphd (changes requested):

package.json
```diff
@@ -46,14 +47,17 @@
   },
   "dependencies": {
     "es6-promisify": "^6.1.0",
-    "minimatch": "^3.0.4"
+    "minimatch": "^3.0.4",
+    "node-polyfill-webpack-plugin": "^1.0.3",
```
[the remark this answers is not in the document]
bcheidemann: Yeah, I'll move those over 👍

tests/webpack.config.js
bcheidemann: Do you mean aliasing fs, path and Buffer? We can definitely do that but obviously we'd need to refactor the shim tests to require fs, path and Buffer instead of the shims so those tests would have to be separated from the rest of the tests. I think that makes sense though as they're webpack specific.

humphd: Right, that's what I was thinking. Given that there are a few ways to do this, having adequate test coverage of all of them is my goal.

bcheidemann: Hey @humphd, I'm still working on this PR but found an issue with the way we're shimming the path module. In short, the alias messes up the import of the path module in filer (which I'd previously overlooked) and as such the path module ends up missing a lot of the methods (all the ones not explicitly replaced by filer). It should be possible to write a webpack plugin which resolves the path module correctly. Are you happy for me to go ahead with this and to make a separate PR for it?

humphd: @bcheidemann I think that sounds like a good idea.  Thanks for pushing on this code a bit, it's great that testing is already discovering ways we can make it better.

bcheidemann: No problem! I'm happy to do it and it's teaching me a lot :D I just wish I had more time to work on it at the moment!

humphd:
Ping me when you want reviews on these.

bcheidemann:
No problem 👍

bcheidemann:
> Ping me when you want reviews on these.

I will come back to these after #771 is merged as I think it makes sense to use the webpack plugin to build these tests. Also, if we move the plugin to it's own repo (as discussed in #771) would you want to move these tests to that repo or leave them here?

humphd:
I think doing it in this repo is probably the best idea, since it has CI setup already, and is so closely related to the other code.  I'm more wondering about reducing the bundle size for Filer for cases where people aren't using the webpack plugin.  We could write a separate set of build/test scripts to be run with `npm-run-all` or something that builds it beside vs. as part of Filer.

package.json
```diff
@@ -46,14 +47,17 @@
   },
   "dependencies": {
     "es6-promisify": "^6.1.0",
+    "path": "^0.12.7",
```
humphd, line 50: Remove this, it's a node built-in module.

humphd (changes requested):
One nit and the lock file needs to be regenerated, otherwise looks good.

PR #771, opened 2021-04-04
Filer Webpack Plugin

### LINKED ISSUE #772: path module not shimmed correctly
When using webpack to shim the path module as described in the documentation, path.isAbsolute and many other path functions are undefined.

bcheidemann:
Closes #772

TODO:
- [x] Write tests for the plugin
- [x] Manual tests & run webpack output in browser to make sure it all works properly


bcheidemann:
@humphd while I was working on this I saw that the Buffer exported in filer seems to be the global Buffer object. I'm wondering therefore if I should remove it from the plugin as it would just be shimming it with itself. Maybe I missing something here?

humphd:
I think that's fine, it will use https://github.com/webpack/node-libs-browser already in our builds, so webpack should do the same?

bcheidemann:
@humphd As discussed I have removed the option to shim buffer and added a brief comment to the README to explain why buffer is not shimmed by the FilerWebpackPlugin and how to handle it. I've also written tests for the FilerWebpackPlugin so this should be more or less ready to review - though I haven't done a full end to end test using it in a test project so there's still a possibility that I will add some more commits if I find any issues (though I think this is unlikely). Either way, I'll let you know when I've finished the end to end tests.

src/webpack-plugin/plugin.js
```diff
@@ -0,0 +1,81 @@
+var path = require('path');
```
humphd, line 1: nit: I would rename this file to `index.js` so you can `require('./webpack-plugin'),` in `src/index.js` above, unless there's a strong reason to keep this name due to some webpack convention?

package.json
```diff
@@ -46,7 +46,9 @@
   },
   "dependencies": {
     "es6-promisify": "^6.1.0",
-    "minimatch": "^3.0.4"
+    "minimatch": "^3.0.4",
+    "path": "^0.12.7",
```
humphd, line 50: Why are we adding `path` vs using node's or the bundler's equivalent?

tests/spec/webpack-plugin/webpack-plugin.spec.js
```diff
@@ -0,0 +1,528 @@
...
+
+    expect(plugin.options.filerDir).to.equal('/node_modules/filer');
+    expect(plugin.options.shimsDir).to.equal('/node_modules/filer/shims');
+    expect(plugin.options.fsProviderDir).to.equal('/node_modules/filer/shims/providers');
+    expect(plugin.options.shimFs).to.equal(true);
+    expect(plugin.options.shimPath).to.equal(true);
+    expect(plugin.options.fsProvider).to.equal('default');
+  });
+
+  it('should instantiate the plugin with invalid options', () => {
```
humphd, line 80: `should throw if provided invalid options` is more clear I think

humphd (changes requested):
I think this looks really good, great work.  The docs and tests are appreciated.  I agree that adding an e2e test that uses this somehow would be good in a follow-up.

I'm a little leery of adding these new dependencies, and further bloating the final bundle.  We could split this out into a separate repo under the `filerjs` org, or could add a second build step to build the webpack plugin separately from Filer, making it an optional extra thing one could import on demand.  Either of those could happen as a follow-up optimization vs. blocking this further.

I'm inclined to take this once you finish the fixes, and we iterate from there.

src/webpack-plugin/plugin.js
bcheidemann: There's no reason to keep it like this as far as I know so I'm happy to change it to index.he

tests/spec/webpack-plugin/webpack-plugin.spec.js
bcheidemann: Yeah, good shout 👍

package.json
bcheidemann: My bad, removed it now

tests/spec/webpack-plugin/webpack-plugin.spec.js
bcheidemann: Done

bcheidemann:
@humphd I patched a couple of minor issues that came up when I was testing the plugin in a test project but if you're happy with those commits then this should be good to go.

bcheidemann:
> I think this looks really good, great work. The docs and tests are appreciated. I agree that adding an e2e test that uses this somehow would be good in a follow-up.
> 
> I'm a little leery of adding these new dependencies, and further bloating the final bundle. We could split this out into a separate repo under the `filerjs` org, or could add a second build step to build the webpack plugin separately from Filer, making it an optional extra thing one could import on demand. Either of those could happen as a follow-up optimization vs. blocking this further.
> 
> I'm inclined to take this once you finish the fixes, and we iterate from there.

I'd be happy to move this out into a separate repo. If we go down that route then I will wait to add full e2e tests for the plugin until it's been moved to the other repo if that's okay.

bcheidemann:
Hey @humphd,

Been super busy these past few weeks so not had time to do anything on filer. Just come back now as I have a few hours free and don't want open PRs etc to go stale. Was wondering if there was a reason why this wasn't merged yet? Happy to make any changes needed =)

All the best,

Ben

humphd:
Hey, apologies for the delay.  I'll merge now.

Issue #773, opened 2021-04-10
BUG: unexpected failures when calling fs functions (e.g. writeFile) with empty options object


bcheidemann:
The following code:
```javascript
fs.writeFile('/a', 'trololol', {}, (err) => {
    // Do something
 });
```
Throws the error:
`
Error: Uncaught TypeError: buffer.copy is not a function (http://localhost:1234/tests.e31bb0bc.js:17358)
    at commonjsGlobal.onerror (mocha.947ba722.js:32655:12)
`

There are some more examples of issues like this but I don't have them to hand at present. I will add them to this bug report when I am able to.

EDIT: I found one similar(ish) issue with fs.readdir. I've opened a separate issue (#775) for this as it appears to be an unrelated issue on initial inspection.

bcheidemann:
I've created a PR with a failing regression test to illustrate the issue ( PR #774 )

humphd:
Nice find.  I think this is a bug in `validate_file_options()`, or how we deal with the encoding being missing on the `options` object:

https://github.com/filerjs/filer/blob/75f2a70c7b8111987f0df6c9105c1b6c65bef3c4/src/filesystem/implementation.js#L1624-L1633

We are returning back the `options` object untouched, which is probably OK (node does the same), but we fail to then use `'utf8'` as our default, like node does in https://github.com/nodejs/node/blob/24fd791184ad0ba27f92ac49b820aec78e491296/lib/fs.js#L1512.  I'll comment more in the PR.

bcheidemann:
Thanks for the pointer - saved me a lot of time!

humphd:
Fixed by #774.

PR #774, opened 2021-04-10
Regression test and fix for issue #773

bcheidemann:
Adds a test which currently fails, demonstrating issue #773

### DIFF
--- a/src/filesystem/implementation.js
+++ b/src/filesystem/implementation.js
@@ -1869,11 +1869,11 @@ function writeFile(context, path, data, options, callback) {
     return callback(new Errors.EINVAL('flags is not valid', path));
   }
 
-  data = data || '';
   if(!Buffer.isBuffer(data)) {
     if(typeof data === 'number') {
       data = '' + data;
     }
+    data = data || '';
     if(typeof data !== 'string') {
       data = Buffer.from(data.toString());
     }
--- a/src/filesystem/implementation.js
+++ b/src/filesystem/implementation.js
@@ -1873,7 +1873,7 @@ function writeFile(context, path, data, options, callback) {
   if(typeof data === 'number') {
     data = '' + data;
   }
-  if(typeof data === 'string' && options.encoding === 'utf8') {
+  if(typeof data === 'string' && (options.encoding || 'utf8') === 'utf8') {
     data = Buffer.from(data);
   }
 
--- a/src/filesystem/implementation.js
+++ b/src/filesystem/implementation.js
@@ -1870,11 +1870,16 @@ function writeFile(context, path, data, options, callback) {
   }
 
   data = data || '';
-  if(typeof data === 'number') {
-    data = '' + data;
-  }
-  if(typeof data === 'string' && (options.encoding || 'utf8') === 'utf8') {
-    data = Buffer.from(data);
+  if(!Buffer.isBuffer(data)) {
+    if(typeof data === 'number') {
+      data = '' + data;
+    }
+    if(typeof data !== 'string') {
+      data = Buffer.from(data.toString());
+    }
+    else {
+      data = Buffer.from(data, options.encoding || 'utf8');
+    }
   }
 
   open_file(context, path, flags, function(err, fileNode) {
--- a/tests/bugs/issue773.js
+++ b/tests/bugs/issue773.js
@@ -1,7 +1,5 @@
-
 'use strict';
 const util = require('../lib/test-utils.js');
-const expect = require('chai').expect;
 
 describe('unexpected failures when calling fs functions (e.g. writeFile) with empty options object, issue773', function() {
   beforeEach(util.setup);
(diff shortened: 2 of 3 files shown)


tests/bugs/issue773.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+const util = require('../lib/test-utils.js');
+
+describe('unexpected failures when calling fs functions (e.g. writeFile) with empty options object, issue773', function() {
+  beforeEach(util.setup);
+  afterEach(util.cleanup);
+
+  it('should call fs.writeFile with an empty options object', function(done) {
+    const fs = util.fs();
+    fs.writeFile('/a', 'trololol', {}, (err) => {
```
humphd, line 10: You can simplify this to:

```js
fs.writeFile('/a', 'trololol', {}, done);
```

humphd (changes requested):
Let's fix this while you're adding this test.

I think the easiest thing would be to [mimic what node does here](https://github.com/nodejs/node/blob/24fd791184ad0ba27f92ac49b820aec78e491296/lib/fs.js#L1512):

```diff
diff --git a/src/filesystem/implementation.js b/src/filesystem/implementation.js
index 3027688..67bc8c0 100644
--- a/src/filesystem/implementation.js
+++ b/src/filesystem/implementation.js
@@ -1873,8 +1873,8 @@ function writeFile(context, path, data, options, callback) {
   if(typeof data === 'number') {
     data = '' + data;
   }
-  if(typeof data === 'string' && options.encoding === 'utf8') {
-    data = Buffer.from(data);
+  if(typeof data === 'string') {
+    data = Buffer.from(data, options.encoding || 'utf8');
   }

   open_file(context, path, flags, function(err, fileNode) {
```

Node does a [bit more](https://github.com/nodejs/node/blob/24fd791184ad0ba27f92ac49b820aec78e491296/lib/internal/fs/utils.js#L327) work to [assert the encoding](https://github.com/nodejs/node/blob/24fd791184ad0ba27f92ac49b820aec78e491296/lib/internal/fs/utils.js#L150-L155), but I don't know if we care beyond making sure we set the default encoding to `'utf8'` if it's missing.

tests/bugs/issue773.js
bcheidemann: Good shout. Will do.

bcheidemann:
> Let's fix this while you're adding this test.
> 
> I think the easiest thing would be to [mimic what node does here](https://github.com/nodejs/node/blob/24fd791184ad0ba27f92ac49b820aec78e491296/lib/fs.js#L1512):
> 
> ```diff
> diff --git a/src/filesystem/implementation.js b/src/filesystem/implementation.js
> index 3027688..67bc8c0 100644
> --- a/src/filesystem/implementation.js
> +++ b/src/filesystem/implementation.js
> @@ -1873,8 +1873,8 @@ function writeFile(context, path, data, options, callback) {
>    if(typeof data === 'number') {
>      data = '' + data;
>    }
> -  if(typeof data === 'string' && options.encoding === 'utf8') {
> -    data = Buffer.from(data);
> +  if(typeof data === 'string') {
> +    data = Buffer.from(data, options.encoding || 'utf8');
>    }
> 
>    open_file(context, path, flags, function(err, fileNode) {
> ```
> 
> Node does a [bit more](https://github.com/nodejs/node/blob/24fd791184ad0ba27f92ac49b820aec78e491296/lib/internal/fs/utils.js#L327) work to [assert the encoding](https://github.com/nodejs/node/blob/24fd791184ad0ba27f92ac49b820aec78e491296/lib/internal/fs/utils.js#L150-L155), but I don't know if we care beyond making sure we set the default encoding to `'utf8'` if it's missing.

I've implemented the equivalent of what node is doing, should be good to review now =)

bcheidemann:
Closes #773

src/filesystem/implementation.js
```diff
@@ -1873,7 +1873,7 @@ function writeFile(context, path, data, options, callback) {
   if(typeof data === 'number') {
     data = '' + data;
   }
-  if(typeof data === 'string' && options.encoding === 'utf8') {
+  if(typeof data === 'string' && (options.encoding || 'utf8') === 'utf8') {
```
humphd, line 1876: This limits us to only ever allowing `'utf8'`, since this doesn't propagate the encoding into `Buffer.from()` as the second arg, and Buffer/node support other encoding types.  What about something like:

```js
if(!Buffer.isBuffer(data)) {
  if(typeof data !== 'string') {
    data = Buffer.from(data.toString());
  } else {
    data = Buffer.from(data, options.encoding || 'utf8');
  }
}
```

Untested code, just thinking.

humphd (changes requested):

src/filesystem/implementation.js
bcheidemann: Yeah, that's better for sure =P I also moved the typeof number check into the !isBuffer if block

src/filesystem/implementation.js
```diff
@@ -1869,12 +1869,17 @@ function writeFile(context, path, data, options, callback) {
     return callback(new Errors.EINVAL('flags is not valid', path));
   }
 
-  data = data || '';
```
bcheidemann, line 1872: @humphd I moved this line of code so that data = 0 would be handled the same as other numbers.

humphd (approved):
Excellent, thank you.

At some point it would be good to do follow-ups in the future that add tests for these new code paths (data being an Object for the `.toString()` call, using other encodings, etc).

Issue #790, opened 2021-11-12
Webpack entry cannot be found after importing filer


thomas-jakemeyn:
# Context

I am trying to add `FilerWebpackPlugin` to an existing Webpack configuration.

# Problem

As soon as I `require('filer')`, Webpack starts failing because it cannot find the entry file anymore.

# Analysis

This problem seems to be caused by a side effect in `src/path.js`:

```js
/**
 * Patch process to add process.cwd(), always giving the root dir.
 * NOTE: this line needs to happen *before* we require in `path`.
 */
process.cwd = () => '/';
```

# Workaround

The only workaround that I found is to save `process.cwd` before importing `filer`, and to restore it afterwards.

```js
const cwd = process.cwd;
const filer = require('filer');
process.cwd = cwd;
```

With this fix, Webpack is able to find the entry file again.

humphd:
cc @bcheidemann, who might have thoughts.

bcheidemann:
Thanks for raising this issue @thomas-jakemeyn.

The problem seems to be that we import the whole of filer when we only need the `FilerWebpackPlugin` class. In hindsight, this is obviously not the best idea! I will put up a PR to deprecate accessing the plugin through the "normal" filer import and add a specific import path for the plugin. @humphd what do you think about:

```javascript
const { FilerWebackPlugin } = require('filer/webpack');
```

@thomas-jakemeyn Until the fix is released, you should be able to import the class directly:

```javascript
// Direct import
const { FilerWebpackPlugin } = require('filer/src/webpack-plugin');
```

Let me know if this works for you! :)

thomas-jakemeyn:
Hello @bcheidemann,

I confirm that using a direct import as you suggested works. Thank you for that!
However, don't you think that the real problem is the side-effect on `process.cwd`?

bcheidemann:
> Hello @bcheidemann,
> 
> I confirm that using a direct import as you suggested works. Thank you for that! However, don't you think that the real problem is the side-effect on `process.cwd`?

@thomas-jakemeyn The `process.cwd` side effect is, as I understand it, required for filer to work properly in the browser. Maybe @humphd can clarify the exact purpose? Regardless, the Webpack config only needs to load the `FilerWebpackPlugin` class which does not import any part of the main filer library (nor should it). As such, if imported correctly, there should be no side-effects.

Are you having trouble with the `process.cwd` side effect other than in the context of the Webpack config (i.e. at runtime)?

[...]

PR #793, opened 2021-11-13
Fix Issue #790

bcheidemann:
This fixes issue #790 by exposing the `FilerWebpackPlugin` class through it's own import.

```javascript
var { FilerWebpackPlugin } = require('filer/webpack');
```

It also deprecates the following:

```javascript
var { FilerWebpackPlugin } = require('filer');
```

And updates the documentation in [README.md](https://github.com/filerjs/filer/blob/master/README.md) to reflect these changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,17 +63,28 @@ your webpack config:
 
 ```javascript
 // webpack.config.js
-var filer = require('filer');
+var { FilerWebpackPlugin } = require('filer/webpack');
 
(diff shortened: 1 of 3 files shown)


bcheidemann:
@humphd what do you think?

humphd (approved):
This makes sense to me.  @thomas-jakemeyn, do you have any input for your use case?

Issue #791, opened 2021-11-12
Shims are not available in NPM package


thomas-jakemeyn:
# Context

I am trying to add `FilerWebpackPlugin` to an existing Webpack configuration.

# Problem

All imports of `fs` are properly replaced by `node_modules/filer/shims/fs.js`. However, that file cannot be found.

# Analysis

The shims are not made available in the NPM package.

humphd:
I've shipped v1.3.1 to npm, do you want to try that? https://github.com/filerjs/filer/releases/tag/v1.3.1

thomas-jakemeyn:
Hello @humphd,

As far as I can see, the shims are still missing in the NPM library v1.3.1.
Also, I quickly checked the history and I could not find any commit that would have fixed this problem.

humphd:
cc @bcheidemann

bcheidemann:
@humphd the "shims" directory in the root of the project needs to be released to NPM. Could you advise on what changes need to be made to facilitate this? Happy to put up a PR.

bcheidemann:
@thomas-jakemeyn this is the workaround until the files are released to NPM:

1. Copy the 'shims' directory from this repo into your project. This can be wherever you want but for arguments sake lets say you copy the contents to `src/filer/shims`.
2. You then need to update the following imports:
```javascript
// src/filer/shims/fs.js
// replace
const { FileSystem } = require('../src/index');
// with
const { FileSystem } = require('filer');

// src/filer/shims/path.js
// replace
const { path } = require('../src/index');
// with
const { path } = require('filer');

// src/filer/shims/providers/default.js
// replace
const { Default } = require('../../src/providers/index');
// with
const { Default } = require('filer/src/providers/index');

// src/filer/shims/providers/indexeddb.js
// replace
const IndexedDB = require('../../src/providers/indexeddb');
// with
const IndexedDB = require('filer/src/providers/indexeddb');

// src/filer/shims/providers/memory.js
// replace
const Memory = require('../../src/providers/memory');
// with
const Memory = require('filer/src/providers/memory');
```
3. Instantiate the `FilerWebpackPlugin` class with the following options
```javascript
// webpack.config.js
module.exports = {
  plugins: [
    new filer.FilerWebpackPlugin({
      shimsDir: '<rootDir>/src/filer/shims`,
    }),
  ],
}
```

[...]

PR #795, opened 2021-11-15
Fix Issue 790 and 791

bcheidemann:
Fixes issues #790 and #791 by including shims and webpack directories in release to NPM

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ module.exports = {
 ---
 **NOTE**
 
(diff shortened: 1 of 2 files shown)


bcheidemann:
@humphd I think this should do it. No need for build scripts for shims or webpack plugin imo. What do you think?

humphd (approved):

humphd:
Awesome, thank you.  I'll land this and do another release

humphd:
https://github.com/filerjs/filer/releases/tag/v1.4.1

